PR #27, opened 2020-04-27
Add the `ListNetworkConnections` action

### LINKED ISSUE #9: Implement the `ListNetworkConnections` action
RRG should support the [`ListNetworkConnections`][network-action] action. It accepts [`ListNetworkConnectionsArgs`][network-request] as input and should respond with [`NetworkConnection`][network-response].

[network-action]: https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/client/grr_response_client/client_actions/network.py#L54-L61
[network-request]: https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/proto/grr_response_proto/flows.proto#L2475-L2482
[network-response]: https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/proto/grr_response_proto/sysinfo.proto#L131-L173

alex65536:
This pull request implements `ListNetworkConnections` action. Fixes #9.

### DIFF
--- a/src/action/network.rs
+++ b/src/action/network.rs
@@ -69,9 +69,9 @@ struct ProcessInfo {
 }
 
 /// A response type for the network connections action.
-pub struct Response<'a> {
+pub struct Response {
     /// Information about the socket which is used by connection.
-    socket_info: &'a ProtocolSocketInfo,
+    socket_info: ProtocolSocketInfo,
     /// Information about the process that owns the connection.
     process_info: Option<ProcessInfo>,
 }
@@ -164,7 +164,7 @@ impl super::Request for Request {
     }
 }
 
-impl<'a> super::Response for Response<'a> {
+impl super::Response for Response {
 
     const RDF_NAME: Option<&'static str> = Some("NetworkConnection");
 
@@ -172,7 +172,7 @@ impl<'a> super::Response for Response<'a> {
 
     fn into_proto(self) -> Self::Proto {
         let mut result: NetworkConnection;
-        result = make_connection_from_socket_info(self.socket_info);
+        result = make_connection_from_socket_info(&self.socket_info);
         if let Some(process_info) = self.process_info {
             result.pid = Some(process_info.pid);
             result.process_name = process_info.name;
@@ -272,7 +272,7 @@ where
             // because the latter lists only the connections associated with a
             // process.
             session.reply(Response {
-                socket_info: &socket_info,
+                socket_info: socket_info,
                 process_info: None,
             })?;
             continue;
@@ -287,7 +287,7 @@ where
         // of them.
         for pid in pids {
             session.reply(Response {
-                socket_info: &socket_info,
+                socket_info: socket_info.clone(),
                 process_info: Some(ProcessInfo::from_system(&system, *pid)),
             })?;
         }
--- a/src/action/network.rs
+++ b/src/action/network.rs
@@ -391,7 +391,7 @@ mod tests {
         }
 
         // We don't check PID for `connection_resp`, because it is unset at
-        // least on Linux and I don't know whether it's set on other systems.
+        // least on Linux and it's unclear whether PID is set on other systems.
         assert_eq!(listen_resp.process_info.as_ref().unwrap().pid, our_pid);
 
         let connection_socket = extract_tcp_info(&connection_resp);
--- a/src/action/network.rs
+++ b/src/action/network.rs
@@ -216,8 +216,18 @@ pub fn handle<S: Session>(
     let mut system = System::new();
     system.refresh_processes();
 
+    let addr_family_flags = netstat2::AddressFamilyFlags::all();
+    // Optimization: when we are asked only for listening connections,
+    // we need only TCP protocol.
+    let protocol_flags = if request.listening_only {
+        netstat2::ProtocolFlags::TCP
+    } else {
+        netstat2::ProtocolFlags::all()
+    };
+
     let connection_iter = netstat2::iterate_sockets_info(
-        netstat2::AddressFamilyFlags::all(), netstat2::ProtocolFlags::all()
+        addr_family_flags,
+        protocol_flags
     );
     let connection_iter = match connection_iter {
         Ok(val) => val,
--- a/src/action/network.rs
+++ b/src/action/network.rs
@@ -37,8 +37,11 @@ impl std::error::Error for Error {
 impl Display for Error {
 
     fn fmt(&self, fmt: &mut Formatter) -> std::fmt::Result {
-        write!(fmt, "failed to list network connections: {}",
-               self.connection_info_error)
+        write!(
+            fmt,
+            "failed to list network connections: {}",
+            self.connection_info_error
+        )
     }
 }
 
@@ -62,8 +65,6 @@ pub struct Response<'a> {
     socket_info: &'a ProtocolSocketInfo,
     process_info: Option<ProcessInfo>
 }
-
-struct SocketInfoWrapper<'a>(&'a ProtocolSocketInfo);
 
 fn make_family(addr: &IpAddr) -> Family {
     match addr {
@@ -97,6 +98,38 @@ fn make_state(state: &TcpState) -> State {
     }
 }
 
+fn make_connection_from_socket_info<'a>(
+    socket_info: &'a ProtocolSocketInfo
+) -> NetworkConnection {
+    match socket_info {
+        Tcp(tcp_info) => NetworkConnection {
+            family: Some(make_family(&tcp_info.local_addr) as i32),
+            r#type: Some(Type::SockStream as i32),
+            local_address: Some(make_network_endpoint(
+                &tcp_info.local_addr,
+                tcp_info.local_port
+            )),
+            remote_address: Some(make_network_endpoint(
+                &tcp_info.remote_addr,
+                tcp_info.remote_port
+            )),
+            state: Some(make_state(&tcp_info.state) as i32),
+            ..Default::default()
+        },
+        Udp(udp_info) => NetworkConnection {
+            family: Some(make_family(&udp_info.local_addr) as i32),
+            r#type: Some(Type::SockDgram as i32),
+            local_address: Some(make_network_endpoint(
+                &udp_info.local_addr,
+                udp_info.local_port
+            )),
+            remote_address: None,
+            state: None,
+            ..Default::default()
+        }
+    }
+}
+
 impl super::Request for Request {
 
     type Proto = ListNetworkConnectionsArgs;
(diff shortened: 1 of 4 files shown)


alex65536:
@googlebot I signed it!

src/action/network.rs
```diff
@@ -0,0 +1,296 @@
...
+        netstat2::ProtocolFlags::all()
+    };
+
+    let connection_iter = netstat2::iterate_sockets_info(
+        addr_family_flags,
+        protocol_flags
+    );
+    let connection_iter = match connection_iter {
+        Ok(val) => val,
+        Err(err) => return Err(session::Error::action(Error::from(err)))
```
panhania, line 239: I think if you implement `Into<session::Error>` for your error type like in the `startup` module, this would be simpler.

src/action/network.rs
```diff
@@ -0,0 +1,296 @@
...
+    };
+
+    for connection in connection_iter {
+        let connection = match connection {
+            Ok(val) => val,
+            Err(err) => {
+                error!("Unable to get socket information: {}", err);
```
panhania, line 246: So far I (and I think other projects as well) use lowercase sentences for log messages.

src/action/network.rs
```diff
@@ -0,0 +1,296 @@
...
+                continue;
+            }
+        };
+        let socket_info = connection.protocol_socket_info;
+
+        if request.listening_only {
+            let tcp_info = match &socket_info {
+                Tcp(tcp_info) => tcp_info,
+                Udp(_) => {
+                    // we are interested only in TCP connections
```
panhania, line 256: For comments though, I prefer uppercase sentences ended with a dot :D

src/action/network.rs
```diff
@@ -0,0 +1,296 @@
...
+                    continue;
+                }
+            };
+            if tcp_info.state != TcpState::Listen {
+                // we are interested only in listening connections
```
panhania, line 261: Ditto.

src/action/network.rs
```diff
@@ -0,0 +1,296 @@
...
+    fn from_system<S: SystemExt>(system: &S, pid: u32) -> ProcessInfo {
+        match system.get_process(pid as i32) {
+            Some(process) => ProcessInfo::from(process),
+            None => ProcessInfo::from_pid(pid)
+        }
+    }
+}
+
+/// Handles requests for the network connection action.
+pub fn handle<S: Session>(
```
panhania, line 217: If you use the `where` syntax for trait bounds, you can make the signature fit in 80 characters (with `where` clause on the next line), like in [`src/action/mod.rs`][1].

[1]: https://github.com/google/rrg/blob/9293f63fff68927e5a045154e40c727236243f66/src/action/mod.rs#L47-L50

src/action/network.rs
```diff
@@ -0,0 +1,296 @@
...
+//! and target IP addresses, the process that owns the connection)
+
+use std::fmt::{Display, Formatter};
+use std::net::IpAddr;
+
+use log::error;
+use sysinfo::{System, SystemExt, Process, ProcessExt};
+use netstat2::{
+    self,
+    ProtocolSocketInfo::{self, Tcp, Udp},
```
panhania, line 19: `Tcp` and `Udp` are used only for pattern matching in one place. In such cases I prefer to do a local import to avoid cluttering the global namespace.

src/action/network.rs
```diff
@@ -0,0 +1,296 @@
...
+    }
+}
+
+/// Converts [netstat2::TcpState][tcp_state] to `State`
+/// enum used in the protobuf
+///
+/// [tcp_state]: ../../../netstat2/enum.TcpState.html
+fn make_state(state: &TcpState) -> State {
+    match state {
+        TcpState::Unknown     => State::Unknown,
```
panhania, line 104: These aligned arrows look nice, but they are bad for diffs—if a new enum value is introduced with a longer name, you have to re-align everything, which is not great. For this reason I prefer not to do them.

src/action/network.rs
```diff
@@ -0,0 +1,296 @@
...
+    NetworkConnection,
+    NetworkEndpoint,
+    network_connection::{Family, Type, State}
+};
+
+use crate::session::{self, Session};
+
+#[derive(Debug)]
+struct Error {
+    connection_info_error: netstat2::error::Error,
```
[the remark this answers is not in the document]
alex65536: Is using a tuple struct, like
```rust
struct Error(netstat2::error::Error)
```
a good option here?

src/action/network.rs
```diff
@@ -0,0 +1,297 @@
...
+    ListNetworkConnectionsArgs,
+    NetworkConnection,
+    NetworkEndpoint,
+    network_connection::{Family, Type, State},
+};
+
+use crate::session::{self, Session};
+
+#[derive(Debug)]
+struct Error(netstat2::error::Error);
```
panhania, line 28: I know this type is not exposed, but I still prefer to put documentation to everything.

src/action/network.rs
```diff
@@ -0,0 +1,297 @@
...
+    }
+}
+
+/// Creates `NetworkConnection` protobuf message from
+/// [netstat2::ProtocolSocketInfo][protocol_socket_info]. The fields that are
+/// impossible to fill using the socket information are set to `None`.
+///
+/// [protocol_socket_info]: ../../../netstat2/enum.ProtocolSocketInfo.html
+fn make_connection_from_socket_info<'a>(
+    socket_info: &'a ProtocolSocketInfo
```
panhania, line 123: Do you need explicit lifetime annotation? I think lifetime inference should handle this easily.

src/action/network.rs
```diff
@@ -0,0 +1,297 @@
...
+        netstat2::ProtocolFlags::TCP
+    } else {
+        netstat2::ProtocolFlags::all()
+    };
+
+    let connection_iter = netstat2::iterate_sockets_info(
+        addr_family_flags,
+        protocol_flags
+    );
+    let connection_iter = match connection_iter {
```
panhania, line 236: Isn't `connection_iter?` enough? Or, if `From`/`Into` doesn't work transitively, something like `connection_iter.map_err(Error::from)?` should work for sure.

alex65536: AFAIK, conversions in Rust are not transitive, and the `?` operator searches only for direct way to cast an error.

alex65536: Also, it seems that `From` is required to implement in order to use use `?`. `Into` is a different trait, and we cannot use it to convert errors. The best thing I can do is to write
```rust
connection_iter.map_err(|err| Error::from(err).into())?;
```
Or should I implement `From<Error>` for `session::Error`?

src/action/network.rs
alex65536: I didn't document the `Error`, because I looked at the example action and saw that `Error` is not documented there. So, is it OK to leave this type undocumented, or is it a bug that `Error` in `actions/startup.rs` is not documented?

panhania: Oh, thanks for pointing this out. I would not call it a "bug", but it is definitely an overlook on my part—the documentation should be there.

src/action/network.rs
panhania: > AFAIK, conversions in Rust are not transitive, and the ? operator searches only for direct way to cast an error.

Looks like you are right.

> Also, it seems that From is required to implement in order to use use ?.

Yep, the [specification](https://doc.rust-lang.org/reference/expressions/operator-expr.html#the-question-mark-operator) seems to confirm that.

Then I guess what you have is fine, although the one with `map_err` does not look terrible either (but might require a line break, so I think the current approach is better).

src/action/network.rs
```diff
@@ -0,0 +1,297 @@
...
+        ProcessInfo {
+            pid: process.pid() as u32,
+            name: Some(process.name().to_string()),
+        }
+    }
+}
+
+impl ProcessInfo {
+
+    /// Constucts a ProcessInfo system from a given process ID. The process
```
panhania, line 196: Missing backticks to denote `ProcessInfo` as an identifier.

Also, I would move the next sentence to next paragraph. I tend to reserve first paragraph for short, single-sentence summary (because it is what is visible on the list of functions in the generated documentation), while leaving the details for the following paragraphs.

src/action/network.rs
```diff
@@ -0,0 +1,297 @@
...
+    /// name will not be set.
+    fn from_pid(pid: u32) -> ProcessInfo {
+        ProcessInfo {
+            pid,
+            name: None,
+        }
+    }
+
+    /// Constucts a ProcessInfo system from a given process ID. The process
```
panhania, line 205: Ditto.

src/action/network.rs
```diff
@@ -0,0 +1,297 @@
...
+            Some(process) => ProcessInfo::from(process),
+            None => ProcessInfo::from_pid(pid),
+        }
+    }
+}
+
+/// Handles requests for the network connection action.
+pub fn handle<S>(session: &mut S, request: Request) -> session::Result<()>
+where
+    S: Session
```
panhania, line 218: Missing trailing comma.

src/action/network.rs
panhania: Addressed in d9700b1fc257127cee293d6350b80c1151095342.

src/action/network.rs
```diff
@@ -0,0 +1,477 @@
...
+
+    use super::*;
+    use netstat2::TcpSocketInfo;
+    use std::net::{TcpStream, TcpListener, UdpSocket, SocketAddr};
+
+    /// Returns all the responses whose `local_addr` is equal to `addr`.
+    fn find_responses_by_addr<'a>(
+        session: &'a session::test::Fake,
+        addr: SocketAddr,
+    ) -> Vec<&'a Response> {
```
Fedya001, line 317: It seems like this explicit lifetime specifier is redundant here. Initially each parameter that is a reference gets its own lifetime parameter. And then compiler applies second lifetime elision rule here. 
Quote from the book: "The second rule is if there is exactly one input lifetime parameter, that lifetime is assigned to all output lifetime parameters: `fn foo<'a>(x: &'a i32) -> &'a i32.`"

alex65536: Yes, you are right :) Thanks, it's fixed now.

src/action/network.rs
```diff
@@ -299,3 +302,178 @@ where
...
+            listen_resp = server_resp[0];
+            connection_resp = server_resp[1];
+        } else {
+            assert_eq!(get_tcp_state(&server_resp[1]), TcpState::Listen);
+            listen_resp = server_resp[1];
+            connection_resp = server_resp[0];
+        }
+
+        // We don't check PID for `connection_resp`, because it is unset at
+        // least on Linux and I don't know whether it's set on other systems.
```
panhania, line 394: The "_I don't know..._" part looks a bit awkward and later it might be hard to say who this "_I_" is. It's better to phrase like "_it's unclear whether PID is set..._" sounds less personal.

alex65536:
> Except for one small comment about a comment, looks good! Once that thing is resolved, I think we can merge this.

This issue is fixed now, so I think this pull request is ready for merge.

PR #28, opened 2020-04-28
Add the `Timeline` action

### LINKED ISSUE #24: Implement the `Timeline` action
RRG should support the [`Timeline`][timeline-action] action. It accepts [`TimelineArgs`][timeline-request] as input and should respond with [`TimelineResult`][timeline-response].

[timeline-action]: https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/client/grr_response_client/client_actions/timeline.py#L20-L42
[timeline-request]: https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/proto/grr_response_proto/timeline.proto#L5-L12
[timeline-response]:
https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/proto/grr_response_proto/timeline.proto#L15-L22

xtsm:
Fixes #24.


src/action/timeline.rs
```diff
@@ -0,0 +1,168 @@
...
+
+    Ok(())
+}
+
+impl super::Request for Request {
+
+    type Proto = rrg_proto::TimelineArgs;
+
+    fn from_proto(proto: rrg_proto::TimelineArgs) -> Request {
+        // TODO fix unwrap
```
alex65536, line 126: It seems that you forgot to fix unwrap :)

xtsm: nope, just waiting for the return type to be changed in master

src/action/timeline.rs
```diff
@@ -0,0 +1,168 @@
...
+        Request {
+            root: PathBuf::from(String::from_utf8(proto.root.unwrap()).unwrap())
```
panhania, line 128: What if the file system does not use UTF-8 encoding (which is very often the case)? For this very reason the proto message uses `bytes` for representing paths, rather than `string` (which requires to be a valid UTF-8).

src/action/timeline.rs
```diff
@@ -0,0 +1,178 @@
...
+use std::path::PathBuf;
+use std::fs;
+use std::fmt::{Display, Formatter};
+use std::os::unix::{fs::MetadataExt, ffi::OsStringExt};
+use sha2::{Digest, Sha256};
+use crate::session::{self, Session, Error, ParseError};
+use crate::gzchunked::GzChunked;
+
+/// A request type for the timeline action.
+pub struct Request {
```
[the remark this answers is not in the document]
xtsm: yes, this one does, it is used in the handle() as an argument

src/action/timeline.rs
```diff
@@ -166,12 +174,66 @@ impl super::Response for ChunkResponse {
...
+    fn into_proto(self) -> DataBlob {
+        DataBlob {
             data: Some(self.data),
             ..Default::default()
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
```
panhania, line 188: This test suite is not very exhaustive. The timeline action is actually one of the few actions there are very easy to test, including all corner cases (e.g. empty directories, deeply nested file hierarchies, files of different types, symlinks with cycles, hardlinks etc.).

In general, there should be _a lot_ of test cases for this action, you can get really wild here.

src/action/timeline.rs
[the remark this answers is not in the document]
panhania: Oh, right. Resolved.

src/action/timeline.rs
```diff
@@ -103,19 +107,33 @@ impl RecurseState {
...
-        };
-        let entry = entry_from_metadata(&metadata, root);
-        self.process_entry(entry, session)?;
-        if metadata.is_dir() && metadata.dev() == self.device {
-            let entry_iter = match read_dir(root) {
-                Ok(iter) => iter,
-                Err(_) => return Ok(()),
+        let mut path = root.clone();
+        let mut dir_iter_stack = Vec::new();
+        loop {
```
panhania, line 112: Is there a reason why you decided to switch from normal recursion to implementing recursion stack yourself? I think the previous code read better and felt more natural.

src/action/timeline.rs
```diff
@@ -188,52 +206,273 @@ impl super::Response for ChunkResponse {
 mod tests {
 
     use super::*;
+    use std::fs::{hard_link, create_dir, write, set_permissions, Permissions};
+    use std::ffi::OsStr;
+    use std::path::Path;
+    use std::os::unix::{ffi::OsStrExt, fs::{symlink, PermissionsExt}};
```
panhania, line 212: This will not compile on Windows.

src/action/timeline.rs
```diff
@@ -188,52 +206,273 @@ impl super::Response for ChunkResponse {
...
 
+    fn handle_for_path<S>(session: &mut S, path: &Path) -> session::Result<()>
+    where
+        S: Session,
+    {
+        let path_bytes = Vec::from(path.as_os_str().as_bytes());
+        let args_proto = TimelineArgs { root: Some(path_bytes) };
+        let request = Request::from_proto(args_proto).unwrap();
+        handle(session, request)
+    }
```
panhania, line 225: This does not seem correct. If you do this, it feels like you are trying to test too much in tests that are not suited for this. If your are testing your handler, then why go through all this conversion to raw bytes and then conversion back from raw bytes?

src/action/timeline.rs
```diff
@@ -188,52 +206,273 @@ impl super::Response for ChunkResponse {
...
+
+    fn entry_for_path(path: &Path) -> TimelineEntry {
+        let metadata = symlink_metadata(path).unwrap();
+        entry_from_metadata(&metadata, &PathBuf::from(path))
+    }
+
+    fn entries_from_session_response(session: &session::test::Fake) -> Vec<TimelineEntry> {
```
panhania, line 232: These tests are now extremely complicated and super hard to follow. You have a lot of test utilites, that might be even more complicated than the code you are testing itself. How do you ensure that you don't have a bug in this code and it really tests what you want to test? This is one of the reasons why writing complex tests and a lot of test utilities is not a good idea (because you should probably write tests for them as well).

Tests should be extremely straightforward, just obvious to read. You can take a look at the tests of [the Python implementation](https://github.com/google/grr/blob/master/grr/client/grr_response_client/client_actions/timeline_test.py). They just use one two-line long helper for creating files and the rest is just a flat script without any branching and complicated loops (but actually, it should not even have these simple loops...).

src/action/timeline.rs
xtsm: the previous code stackoverflow'd in case of deep directory trees and i had no idea how to fix that without getting rid of recursion

src/action/timeline.rs
xtsm: The Python implementation differs from my Rust one in lots of places, I don't think I can make my tests look as neat as Python ones.

For example, if we look at this specific function (`entries_from_session_response`), we can tell that most of it is just retrieving blocks from fake session transfer store and checking their digests. In Python, there's nothing like that because the FS walk and gzchunked serialization are implemented via coroutines, i.e. completely separated from session logic. I planned to do the exact same thing in the beginning, but you said that coroutines aren't idiomatic to Rust (yet) so I just stuck the whole thing to the session communication to avoid storing a lot of intermediate data.

I _guess_ I can do something about handler and directly test RecurseState instead of handle(), but the chunk unpacking here has to stay.

Or to be moved into fake session logic somehow, but is that how transfer store works in general and not in this specific action?

src/action/timeline.rs
```diff
@@ -46,22 +45,95 @@ struct RecurseState {
...
+    cfg_if! {
+        if #[cfg(target_family = "unix")] {
+            use std::os::unix::ffi::OsStrExt;
+            Ok(Vec::from(s.as_bytes()))
+        } else if #[cfg(target_family = "windows")] {
+            use std::os::windows::ffi::OsStrExt;
+            use byteorder::{LittleEndian, WriteBytesExt};
+            s.encode_wide().try_fold(Vec::new(), |mut stream, ch| {
+                stream.write_u16::<LittleEndian>(ch).map(|_| stream)
+            })
```
panhania, line 74: Interesting, I think this is a tough problem. Is the choice of using little-endian arbitrary or was there some reason for choosing it? I think it warrants a comment or perhaps a link to some resources.

src/action/timeline.rs
```diff
@@ -46,22 +45,95 @@ struct RecurseState {
...
+            OsString::from_vec(Vec::from(bytes))
+        } else if #[cfg(target_family = "windows")] {
+            use std::os::windows::ffi::OsStringExt;
+            use byteorder::{LittleEndian, ReadBytesExt};
+            let mut wchars = Vec::new();
+            let mut bytes_slice = bytes;
+            while let Ok(wchar) = bytes_slice.read_u16::<LittleEndian>() {
+                wchars.push(wchar);
+            }
+            OsString::from_wide(wchars.as_slice())
```
panhania, line 95: Ditto.

src/action/timeline.rs
panhania: Yeah, as the code is right now, it's hard to not have any helpers. Still, I think this `entries_from_session_response` is really all you need, and the rest should be just explicit assertions (instead of using `diff_entries`, loops and alike). Moreover, I don't really understand why you have some assertions in this method. All it should do is to parse all the data that you delivered to the transfer store and return it in form of `TimelineEntry`.

src/action/timeline.rs
xtsm: According to a quick search, UTF-16 that Windows operates with is virtually always little-endian.
Unfortunately, I can't check in practice what does Windows GRR server send to us, so I'm just assuming that it's just regular UTF16-LE without BOM.
But yeah, probably worth a comment.

src/action/timeline.rs
xtsm: We need these asserts because we would want to know if the chunk hashing has gone wrong for some reason. If I understand the whole process correctly, it would mean that the server won't be able to extract the blocks from the transfer store, which is pretty bad.

I also need `diff_entries` because if I remove it, I'll just have to copy its body into every test. I guess I'll just move the final assert out of it and only leave the transformation into diff, that would actually be nicer.

panhania: > I also need diff_entries because if I remove it, I'll just have to copy its body into every test. I guess I'll just move the final assert out of it and only leave the transformation into diff, that would actually be nicer.

But why? You know what the output should be. You literally have a variable named `expected_entries` (which you should not—it even leads to a duplicated functionality, like you sort this vector before `diff_entries` and inside `diff_entries` in most cases and you likely did not notice this precisely because there is just too much logic even for simple tests).

The only test that *could* have some use of it is the one where you generate a deep directory structure. But as you said, the thing that you are testing there is the ability to handle deep directories, not whether the fields have the expected values. So I would argue that even there you don't need it.

xtsm: The `expected_entries` should be unordered because the directory listing order is undefined. I just couldn't find an actual unordered collection that could store protos so I have to compare sorted vectors.

> the thing that you are testing there is the ability to handle deep directories, not whether the fields have the expected values

Aren't those things equivalent?

And, I'm pretty sure that we discussed earlier that diff gives us more information than just assert_eq. Even if we only have 3-4 entries, it still saves us some time because it tells exactly which entries are deviating from expected ones.

src/action/timeline.rs
panhania: This is still missing a comment in the code to explain reasoning.

src/action/timeline.rs
```diff
@@ -0,0 +1,567 @@
...
+
+    const RDF_NAME: Option<&'static str> = Some("DataBlob");
+
+    type Proto = DataBlob;
+
+    fn into_proto(self) -> DataBlob {
+        DataBlob {
+            data: Some(self.data),
+            ..Default::default()
+        }
```
panhania, line 274: You should be able to use `self.data.into()` now.

src/action/timeline.rs
```diff
@@ -0,0 +1,567 @@
...
+            }
+            OsString::from_wide(wchars.as_slice())
+        } else {
+            compile_error!("unsupported OS family");
+        }
+    }
+}
+
+/// Encodes filesystem metadata into timeline entry proto.
+fn entry_from_metadata(metadata: &Metadata, path: &PathBuf) -> std::io::Result<TimelineEntry>
```
panhania, line 103: Passing `PathBuf` by reference is an anti-pattern (same way as `&String` or `&Vec<_>`). Prefer to use `Path` instead.

src/action/timeline.rs
panhania: > The expected_entries should be unordered because the directory listing order is undefined. I just couldn't find an actual unordered collection that could store protos so I have to compare sorted vectors.

I don't see the connection, why ordering or not of the directory listing should affect your expectations? Because directory listing is undefined, in your test case you sort the output. And then it becomes "defined" (for the test sake). And you can make assertions on it.

> And, I'm pretty sure that we discussed earlier that diff gives us more information than just assert_eq. Even if we only have 3-4 entries, it still saves us some time because it tells exactly which entries are deviating from expected ones.

It gives you more information if you don't have explicit assertions on particular fields. If you have, then it's the same, except that you get a very confusing error message that you have to parse yourself and it requires a lot of logic.

src/action/timeline.rs
xtsm: You mean `Vec` into `Some(Vec)`?
Why?

panhania: I mean that now there is a `From<Vec<u8>>` [implemented](https://github.com/google/rrg/blob/cdec0598987acf03df22552223a5e91a0d25c370/proto/src/lib.rs#L38-L46) for `DataBlob`, so you can convert your `Vec<u8>` into `DataBlob` with just a single `.into()` call.

PR #29, opened 2020-04-28
Add the `EnumerateInterfaces` action

### LINKED ISSUE #23: Implement the `EnumerateInterfaces` action
RRG should support the [`EnumerateInterfaces`][enum-action] action. It does not have any input arguments and should respond with [`Interface`][enum-response].

[enum-action]: https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/client/grr_response_client/client_actions/linux/linux.py#L175-L182
[enum-response]:
https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/proto/grr_response_proto/jobs.proto#L940-L986

AlexJokel:
This pull request implements the `EnumerateInterfaces` action, closes #23.

I could not find anything about [these](https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/proto/grr_response_proto/jobs.proto#L969-L986) DHCP `Interface` protobuf parts in both [existing implementation](https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/client/grr_response_client/client_actions/linux/linux.py#L115) and the Rust lib I used ([`pnet`](https://docs.rs/pnet/0.25.0/pnet/index.html)). So I just left them blank.

### DIFF
--- a/src/action/interfaces.rs
+++ b/src/action/interfaces.rs
@@ -8,14 +8,12 @@
 //! The interfaces action lists all network interfaces available for the client,
 //! collecting their names, MAC and IP addresses.
 
-use std::net::IpAddr;
-
 use ipnetwork::IpNetwork;
 use pnet::{
     datalink::{self, NetworkInterface},
     util::MacAddr,
 };
-use rrg_proto::{Interface, NetworkAddress, network_address::Family};
+use rrg_proto::{Interface, NetworkAddress};
 
 use crate::session::{self, Session};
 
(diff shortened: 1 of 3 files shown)


src/action/interfaces.rs
```diff
@@ -0,0 +1,89 @@
+// Copyright 2020 Google LLC
+//
+// Use of this source code is governed by an MIT-style license that can be found
+// in the LICENSE file or at https://opensource.org/licenses/MIT.
+
+//! A handler and associated types for the interfaces action.
+//!
+//! The interfaces action lists all network interfaces available for the client,
```
panhania, line 8: I think *"on the client"* is more correct.

src/action/interfaces.rs
```diff
@@ -87,3 +88,38 @@ impl super::Response for Response {
...
+        assert!(handle(&mut session, ()).is_ok());
+
+        let mut responses =
+            HashSet::<&Response>::with_capacity(session.reply_count());
+        for i in 0..session.reply_count() {
+            let response = &session.reply::<Response>(i);
+            assert!(!responses.contains(response));
+            responses.insert(response);
+        }
+    }
```
panhania, line 124: In my opinion, this test does not provide you with much value. I think it should be removed (along with deriving some questionable traits for the `Response` type). The loopback test looks alright and I think its all we need.

AlexJokel: Okay. Do you deem `Debug` trait questionable here?

src/action/interfaces.rs
```diff
@@ -81,7 +83,16 @@ impl super::Response for Response {
...
+            mac_address: match self.interface.mac {
+                Some(mac) => Some(mac_to_vec(mac)),
+                None => {
+                    error!(
+                        "unable to get MAC address for {} interface",
+                        self.interface.name,
+                    );
+                    None
+                },
+            },
```
panhania, line 95: I would assign result of this `match` to a separate variable. Mixing-in imperative logic (printing errors) into structure construction is generally not a good idea.

src/action/interfaces.rs
panhania: I think `Debug` is harmless and can be safely included in most of the structs.

PR #31, opened 2020-05-04
Implement utilities for the gzchunked format

xtsm:
Gzchunked is a data streaming format used in some actions, e.g. Timeline (#28)


src/gzchunked/mod.rs
```diff
@@ -0,0 +1,43 @@
+// Copyright 2020 Google LLC
+//
+// Use of this source code is governed by an MIT-style license that can be found
+// in the LICENSE file or at https://opensource.org/licenses/MIT.
+
+//! Utils for forming gzchunked streams.
+
```
panhania, line 7: This module has no documentation.

src/gzchunked/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+const BLOCK_SIZE: usize = 10 << 20;
+
+pub struct GzChunked {
+    encoder: GzEncoder<Vec<u8>>,
+}
+
+impl GzChunked {
+    pub fn new() -> GzChunked {
+        GzChunked {
+            encoder: GzEncoder::new(Vec::new(), flate2::Compression::default())
```
panhania, line 21: Prefer to use some specific compression level, rather to delegate this to whatever the library author chooses. Here, the default level is `6`, which I think is a sensible choice (but something like `5` would be better, I think), but if they decide to change it to something bigger at some point, RRG suddenly will become slower. It also hides this information from other developers and they are forced to dive into the library source to figure it out. An explicit value is better.

The timeline action should be first and foremost fast but, if possible, it should not blow up the response size too much either. I think some benchmarking is needed to check what level gives appropriate balance.

src/gzchunked/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+        Ok(Some(self.next_chunk()?))
+    }
+    pub fn next_chunk(&mut self) -> std::io::Result<Vec<u8>> {
+        self.encoder.flush()?;
+        self.encoder.try_finish()?;
+        let ret = Ok(self.encoder.get_ref().clone());
+        self.encoder = GzEncoder::new(Vec::new(), flate2::Compression::default());
+        ret
+    }
+}
```
panhania, line 43: This module is lacking tests. For this, I think you also need to implement a decoder. Which makes sense and should not be too hard.

src/gzchunked/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+        self.encoder.write_all(&(buf.len() as u64).to_be_bytes())?;
+        self.encoder.write_all(buf)?;
+        Ok(())
+    }
+    pub fn try_next_chunk(&mut self) -> std::io::Result<Option<Vec<u8>>> {
+        self.encoder.flush()?;
+        if self.encoder.get_ref().len() < BLOCK_SIZE {
+            return Ok(None)
+        }
+        Ok(Some(self.next_chunk()?))
```
panhania, line 34: Try to put more thought into whitespace placement. Here, I think putting an empty line before `Ok(...)` would improve the readability. Also, there are empty lines missing between function declarations (and after `impl`).

src/gzchunked/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+    }
+
+    /// Retrieves next gzipped block without checking its size.
+    pub fn next_chunk(&mut self) -> std::io::Result<Vec<u8>> {
+        self.encoder.flush()?;
+        self.encoder.try_finish()?;
+
+        let ret = Ok(self.encoder.get_ref().clone());
+        self.encoder = GzEncoder::new(Vec::new(), flate2::Compression::default());
```
panhania, line 58: In the constructor you respect the compression level specified by the caller, but as soon as `.next_chunk()` is used, this specified compression is discarded and the default one is used. I don't think it should be this way.

src/gzchunked/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+
+        Ok(Some(self.next_chunk()?))
+    }
+
+    /// Retrieves next gzipped block without checking its size.
+    pub fn next_chunk(&mut self) -> std::io::Result<Vec<u8>> {
+        self.encoder.flush()?;
+        self.encoder.try_finish()?;
+
+        let ret = Ok(self.encoder.get_ref().clone());
```
panhania, line 57: This `.clone()` is a bit worrying here because it performs an expensive copy. Why do you call `.try_finish()` instead of `.finish()` (that would return an owned value without copying)?

src/gzchunked/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+    pub fn new() -> GzChunkedDecoder {
+        GzChunkedDecoder {
+            queue: VecDeque::new()
+        }
+    }
+
+    /// Decodes next gzchunked block and puts all results into the internal queue.
+    pub fn write(&mut self, buf: &[u8]) -> std::io::Result<()> {
+        let mut decoder = GzDecoder::new(buf);
+        let mut chunked_data_vec : Vec<u8> = Vec::new();
```
panhania, line 74: Redundant space before `:`.

src/gzchunked/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+        decoder.read_to_end(&mut chunked_data_vec)?;
+        let mut chunked_data = chunked_data_vec.as_slice();
+        while !chunked_data.is_empty() {
+            let (length_slice, remainder) = chunked_data.split_at(8);
+            let mut length : [u8; 8] = Default::default();
+            length.copy_from_slice(length_slice);
+            let length = u64::from_be_bytes(length);
+            let (data_slice, remainder) = remainder.split_at(length as usize);
```
panhania, line 82: There is a suspiciously large amount of manual bit-fiddling and usually it is not a good idea. Is this not something that could be rewritten with the idiomatic `Read` and `Write` traits or something like the [`byteorder`](https://docs.rs/byteorder) or the [`bytes`](https://docs.rs/bytes) crates?

Cargo.toml
```diff
@@ -15,3 +15,5 @@ rrg-proto = { path = "proto/" }
 simplelog = { version = "0.7.6" }
 structopt = { version = "0.3.12" }
 sys-info = { version = "0.5.1" }
+flate2 = { version = "1.0.14" }
+rand = { version = "0.7.3" }
```
panhania, line 19: This is a test-only dependency, I believe.

src/gzchunked/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    use super::*;
+    use rand::{Rng, SeedableRng, rngs::StdRng};
+
+    #[test]
+    fn test_encode_and_decode_random() {
+        let mut rng = StdRng::seed_from_u64(20200509);
+        let mut encoder = GzChunkedEncoder::new(Compression::default());
+        let mut decoder = GzChunkedDecoder::new();
+        let mut expected_data : Vec<Vec<u8>> = Vec::new();
```
panhania, line 107: Redundant space before `:` (here and below).

src/gzchunked/mod.rs
```diff
@@ -0,0 +1,128 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+
+    use super::*;
+    use rand::{Rng, SeedableRng, rngs::StdRng};
+
+    #[test]
+    fn test_encode_and_decode_random() {
```
panhania, line 103: I would also add test cases with some corner cases (e.g. empty) and also with non-random data that is easier to follow (the random one is good for testing this with larger amounts of data, but will be hard to debug if something doesn't work).

src/gzchunked/mod.rs
```diff
@@ -99,16 +116,76 @@ mod tests {
...
     use rand::{Rng, SeedableRng, rngs::StdRng};
 
+    #[test]
+    fn test_encode_and_decode_empty() {
+        let mut encoder = GzChunkedEncoder::new(GzChunkedCompression::default());
+        let mut decoder = GzChunkedDecoder::new();
+
+        let encoded_block = encoder.next_chunk().unwrap();
+        // should contain gzip header
+        assert_ne!(encoded_block.len(), 0);
```
panhania, line 126: I think `assert!(!encoded_block.is_empty())` would read better.

src/gzchunked/mod.rs
```diff
@@ -99,16 +116,76 @@ mod tests {
     use super::*;
     use rand::{Rng, SeedableRng, rngs::StdRng};
 
+    #[test]
+    fn test_encode_and_decode_empty() {
+        let mut encoder = GzChunkedEncoder::new(GzChunkedCompression::default());
+        let mut decoder = GzChunkedDecoder::new();
+
+        let encoded_block = encoder.next_chunk().unwrap();
+        // should contain gzip header
```
panhania, line 125: Comments should be proper written text (start with uppercase letter and end with a dot).

src/gzchunked/mod.rs
```diff
@@ -99,16 +116,76 @@ mod tests {
...
+        let blocks = vec![
+            vec![1, 2, 3, 4],
+            vec![],
+            vec![1, 2, 3, 4, 5, 6, 7, 8],
+            vec![1],
+        ];
+
+        for block in &blocks {
+            encoder.write(block.as_slice()).unwrap();
+        }
```
panhania, line 146: As I mentioned, test scenarios should read like linear script and not force the reader to do any kind of evaluation in their heads. Here, instead of using a for loop, it is better to manually write `encoder.write(&[1, 2, 3, 4])` etc.

src/gzchunked/mod.rs
```diff
@@ -99,16 +116,76 @@ mod tests {
...
+
+        let encoded_block = encoder.next_chunk().unwrap();
+        decoder.write(encoded_block.as_slice()).unwrap();
+
+        for block in blocks {
+            assert_eq!(block, decoder.try_next_data().unwrap());
+        }
```
panhania, line 153: Here, it is not only readability issue but also debuggability. If the test fails, you don't really know at which iteration it failed (I guess in this case you could figure it out, but if someone changes the test data to be non-unique, then you have a problem again).

Writing test assertions like this is very tempting if you create variables, but if you follow the advice from the previous comment, then it is just not possible.

src/gzchunked/mod.rs
```diff
@@ -99,16 +116,76 @@ mod tests {
...
+
+        for block in blocks {
+            encoder.write(block.as_slice()).unwrap();
+            let encoded_block = encoder.next_chunk().unwrap();
+
+            decoder.write(encoded_block.as_slice()).unwrap();
+
+            assert_eq!(block, decoder.try_next_data().unwrap());
+            assert!(decoder.try_next_data().is_none());
+        }
```
panhania, line 176: This feels like these should be separate test cases.

xtsm: no, this is supposed to ensure that the same instance of encoder or decoder can correctly process multiple blocks

src/gzchunked/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+        encoder.write(&[1]).unwrap();
+
+        let encoded_block = encoder.next_chunk().unwrap();
+        decoder.write(encoded_block.as_slice()).unwrap();
+
+        assert_eq!([1, 2, 3, 4], decoder.try_next_data().unwrap().as_slice());
+        assert_eq!([] as [u8; 0], decoder.try_next_data().unwrap().as_slice());
+        assert_eq!([1, 2, 3, 4, 5, 6, 7, 8], decoder.try_next_data().unwrap().as_slice());
+        assert_eq!([1], decoder.try_next_data().unwrap().as_slice());
+        assert!(decoder.try_next_data().is_none());
```
panhania, line 151: In Rust (at least in the standard library, which I tend to use as the reference if I am unsure) the expected value is usually specified as the right argument for `assert_eq!`, not the left one.

Also, is there are reason why you do these `.unwrap().as_slice()` rather than comparing with the real value? I would do it like this:

```rust
assert_eq!(decoder.try_next_data(), Some(vec!(1, 2, 3, 4)))`
// (...)
assert_eq!(decoder.try_next_data(), None)
```

I think its both less verbose and clearer what you actually compare against (e.g. explicit `Some`, not an `.unwrap()`). The comparison against `None` with `.is_none()` in general is okay, but if you have a chain of `assert_eq!` before it, I think it just aligns more nicely, if you use explicit comparison against `None`.

Using `vec!` of course causes an extra allocation, but in tests we don't really care and readability should be prioritized.

src/gzchunked/mod.rs
```diff
@@ -0,0 +1,210 @@
...
+            expected_data.push(data);
+        }
+        let last_block = encoder.next_chunk().unwrap();
+        decoder.write(last_block.as_slice()).unwrap();
+        while let Some(data) = decoder.try_next_data() {
+            decoded_data.push(data);
+        }
+        assert_eq!(expected_data, decoded_data);
+    }
+}
```
panhania, line 210: I would also add (very simple) tests that encoding with non-default (lowest and highest, preferably) compression level works as expected—it should, because this just delegates to `flate2` for now, but tests are supposed to guard you against breakages caused by implementation changes.

PR #32, opened 2020-05-14
Add `GetPlatformInfo` action

### LINKED ISSUE #6: Implement the `GetPlatformInfo` action
RRG should support the [`GetPlatformInfo`][info-action] action. It does not have any input arguments and should respond with [`Uname`][info-response].

[info-action]: https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/client/grr_response_client/client_actions/admin.py#L60-L66
[info-response]:
https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/proto/grr_response_proto/jobs.proto#L859-L901

nikkita1267:
This pull request implements the `GetPlatformInfo` action, closes #6.

I left `libc_ver` and `install_date` fields blank because function `Uname.FromCurrentSystem` ([link](https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/core/grr_response_core/lib/rdfvalues/client.py#L536)) doesn't fill them. I also add [libc crate](https://docs.rs/libc/0.2.70/libc/) which provides `uname` system call.

All fields, except `libc_ver` and `install_date`, are currently filled for Linux.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,3 +15,4 @@ rrg-proto = { path = "proto/" }
 simplelog = { version = "0.7.6" }
 structopt = { version = "0.3.12" }
 sys-info = { version = "0.5.1" }
+libc = { version = "0.2" }
--- a/src/action/mod.rs
+++ b/src/action/mod.rs
@@ -5,6 +5,7 @@
 
 pub mod metadata;
 pub mod startup;
+pub mod platform_info;
 
 use crate::session::{self, Session, Task};
 
@@ -51,6 +52,7 @@ where
     match action {
         "SendStartupInfo" => task.execute(self::startup::handle),
         "GetClientInfo" => task.execute(self::metadata::handle),
+        "GetPlatformInfo" => task.execute(self::platform_info::handle),
         action => return Err(session::Error::Dispatch(String::from(action))),
     }
 }
--- a/src/action/mod.rs
+++ b/src/action/mod.rs
@@ -3,20 +3,61 @@
 // Use of this source code is governed by an MIT-style license that can be found
 // in the LICENSE file or at https://opensource.org/licenses/MIT.
 
+//! Handlers and types for agent's actions.
+//!
+//! The basic functionality that a GRR agent exposes is called an _action_.
+//! Actions are invoked by the server (when running a _flow_), should gather
+//! requested information and report back to the server.
+//!
+//! In RRG each action consists of three components: a request type, a response
+//! type and an action handler. Request and response types wrap lower-level
+//! Protocol Buffer messages sent by and to the GRR server. Handlers accept one
+//! instance of the corresponding request type and send some (zero or more)
+//! instances of the corresponding response type.
+
 pub mod metadata;
 pub mod startup;
 pub mod platform_info;
 
 use crate::session::{self, Session, Task};
 
-pub trait Request {
+/// Abstraction for action-specific requests.
+///
+/// Protocol Buffer messages received from the GRR server are not necessarily
+/// easy to work with and are hardly idiomatic to Rust. For this reason, actions
+/// should define more structured data types to represent their input and should
+/// be able to parse raw messages into them.
+pub trait Request: Sized {
+
+    /// A type of the corresponding raw proto message.
     type Proto: prost::Message + Default;
-    fn from_proto(proto: Self::Proto) -> Self;
+
+    /// A method for converting raw proto messages into structured requests.
+    fn from_proto(proto: Self::Proto) -> Result<Self, session::ParseError>;
 }
 
-pub trait Response {
+/// Abstraction for action-specific responses.
+///
+/// Like with the [`Request`] type, Protocol Buffer messages sent to the GRR
+/// server are very idiomatic to Rust. For this reason, actions should define
+/// more structured data types to represent responses and provide a way to
+/// convert them into the wire format.
+///
+/// Note that because of the design flaws in the protocol, actions also need to
+/// specify a name of the wrapper RDF class from the Python implementation.
+/// Hopefully, one day this issue would be fixed and class names will not leak
+/// into the specification.
+///
+/// [`Request`]: trait.Request.html
+pub trait Response: Sized {
+
+    /// A name of the corresponding RDF class.
     const RDF_NAME: Option<&'static str>;
+
+    /// A type of the corresponding raw proto message.
     type Proto: prost::Message + Default;
+
+    /// A method for converting structured responses into raw proto messages.
     fn into_proto(self) -> Self::Proto;
 }
 
@@ -24,7 +65,8 @@ impl Request for () {
 
     type Proto = ();
 
-    fn from_proto(_: ()) {
+    fn from_proto(unit: ()) -> Result<(), session::ParseError> {
+        Ok(unit)
     }
 }
 
--- a/src/action/platform_info.rs
+++ b/src/action/platform_info.rs
@@ -1,0 +1,106 @@
+// Copyright 2020 Google LLC
+//
+// Use of this source code is governed by an MIT-style license that can be found
+// in the LICENSE file or at https://opensource.org/licenses/MIT.
+
+extern crate sys_info;
+extern crate libc;
+// extern crate sysinfo;
+
+use sys_info::{linux_os_release, os_type, hostname};
+use libc::{uname, utsname, c_char};
+use std::ffi::CStr;
+use std::option::Option;
+use crate::session::{self, Session};
+
+use rrg_proto::Uname;
+
+#[derive(Default)]
+pub struct PlatformInfo {
+    system: Option<String>,
+    release_name: Option<String>,
+    version_id: Option<String>,
+    machine: Option<String>,
+    kernel_release: Option<String>,
+    fqdn: Option<String>,
+    architecture: Option<String>,
+    node: Option<String>
+}
+
+pub struct Response {
+    platform_information: PlatformInfo,
+}
+
+#[inline(always)]
+fn convert_raw_string(c_string: &[c_char]) -> String {
+    unsafe { 
+        String::from(CStr::from_ptr(c_string.as_ptr()).to_string_lossy().into_owned())
+    } 
+}
+
+pub fn handle<S: Session>(session:&mut S, _: ()) -> session::Result<()> {
+    let os_type = os_type().expect("Can't get os type");
+    match os_type.as_str() {
+        "Linux" => {
+            let linux_release_info = linux_os_release()
+                                        .expect("Can't get info about linux system");
+            let mut system_info: utsname;
+
+            unsafe {
+                system_info = std::mem::zeroed();
+                uname(&mut system_info);
+            }
+
+            session.reply(Response {
+                platform_information: PlatformInfo {
+                    system: Some(os_type.to_string()),
+                    release_name: linux_release_info.name,
+                    version_id: linux_release_info.version_id,
+                    machine: Some(convert_raw_string(&system_info.machine)),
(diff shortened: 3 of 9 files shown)


nikkita1267:
@googlebot I signed it!

src/action/platform_info.rs
```diff
@@ -0,0 +1,274 @@
+// Copyright 2020 Google LLC
+//
+// Use of this source code is governed by an MIT-style license that can be found
+// in the LICENSE file or at https://opensource.org/licenses/MIT.
+
+extern crate sys_info;
+
+/// Using for `uname` syscall.
+extern crate libc;
```
panhania, line 9: Why do you need these declarations?

src/action/platform_info.rs
```diff
@@ -0,0 +1,274 @@
...
+use rrg_proto::Uname;
+
+#[derive(Debug)]
+enum Error {
+    CannotGetOSType(sys_info::Error),
+    CannotGetLinuxRelease(sys_info::Error)
+}
+
+impl std::error::Error for Error {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
```
panhania, line 27: I tend to put a blank line after `impl`, `trait` and the first item (it looks better for some reason for me).

src/action/platform_info.rs
```diff
@@ -0,0 +1,274 @@
...
+
+impl Display for Error {
+    fn fmt(&self, fmt: &mut Formatter) -> std::fmt::Result {
+        use Error::*;
+
+        match *self {
+            CannotGetOSType(ref error) => {
+                write!(fmt, "can't get OS type error: {}", error)
+            },
+
```
panhania, line 45: Excessive blank line.

Cargo.toml
```diff
@@ -15,3 +15,4 @@ rrg-proto = { path = "proto/" }
 simplelog = { version = "0.7.6" }
 structopt = { version = "0.3.12" }
 sys-info = { version = "0.5.1" }
```
panhania, line 17: Recently, I have made a change to use the `sysinfo` crate rather than `sys-info`. `sys-info` does not seem to work on other operating systems and is more low-level, whereas `sysinfo` gives you more abstractions and works on pretty much all supported platforms and saves us from having a lot of `unsafe` calls. Have you considered using it?

Right now, the code will not compile on Windows and macOS, I believe.

src/action/platform_info.rs
```diff
@@ -0,0 +1,274 @@
...
+    kernel_release: Option<String>,
+    fqdn: Option<String>,
+    architecture: Option<String>,
+    node: Option<String>
+}
+
+/// A Response type for `GetPlatformInfo` action
+pub struct Response {
+    /// Client platform information
+    platform_information: PlatformInfo,
```
panhania, line 75: Is there a reason to have separate types `PlatformInfo` and `Response` (that is specific to an action called `GetPlatformInfo`)?

src/action/platform_info.rs
```diff
@@ -0,0 +1,274 @@
...
+}
+
+/// Funcion that converts raw C-strings into `String` type
+#[inline(always)]
```
panhania, line 79: Any reason for forcing inlinining over letting the compiler decide (or a slightly less "demanding" variant: just `#[inline]`)?

src/action/platform_info.rs
```diff
@@ -0,0 +1,274 @@
...
+#[derive(Default)]
+pub struct PlatformInfo {
+    system: Option<String>,
+    release_name: Option<String>,
+    version_id: Option<String>,
+    machine: Option<String>,
+    kernel_release: Option<String>,
+    fqdn: Option<String>,
+    architecture: Option<String>,
+    node: Option<String>
```
panhania, line 69: It feels like a lot of these variants would benefit from having a dedicated enum rather than using raw strings for everything.

Cargo.toml
nikkita1267: I try to find something useful for my action in crate `sysinfo`, but, seems to me, this package doesn't provide any information about OS, only about system at all.
Also as documentation of [sys-info](https://docs.rs/sys-info/0.7.0/sys_info/) says that it supports Windows and Mac OS too.

src/action/platform_info.rs
nikkita1267: No reason, now fixed:)

src/action/platform_info.rs
nikkita1267: Ok, I fixed that

src/action/platform_info.rs
nikkita1267: Fixed!

src/action/platform_info.rs
nikkita1267: No need now, I removed them.

src/action/platform_info.rs
nikkita1267: I don't quite understand, what do you mean by _dedicated enum_?

src/action/platform_info.rs
nikkita1267: I thought that it would be nice if the conversion code was inlined, but probably this solution should be left to the compiler. Thanks!

src/action/platform_info.rs
```diff
@@ -0,0 +1,271 @@
+// Copyright 2020 Google LLC
+//
+// Use of this source code is governed by an MIT-style license that can be found
+// in the LICENSE file or at https://opensource.org/licenses/MIT.
+
+use sys_info::{linux_os_release, os_type, hostname};
+#[cfg(target_os = "linux")]
+use libc::{uname, utsname};
+use libc::c_char;
```
panhania, line 9: Try to put platform-specific imports locally next to the places that actually need them.

src/action/platform_info.rs
```diff
@@ -0,0 +1,271 @@
...
+
+    const RDF_NAME: Option<&'static str> = Some("Uname");
+
+    type Proto = rrg_proto::Uname;
+
+    /// Convert `Response` struct to protobuf message `Uname`
+    fn into_proto(self) -> rrg_proto::Uname {
+        Uname {
+            system: self.system.clone(),
+            release: self.release_name.clone(),
```
panhania, line 136: Why do you need to clone? `into_proto` takes an owned response.

src/action/platform_info.rs
```diff
@@ -0,0 +1,271 @@
...
+            version: self.version_id,
+            machine: self.machine,
+            kernel: self.kernel_release,
+            fqdn: self.fqdn,
+            architecture: self.architecture.clone(),
```
panhania, line 141: Ditto.

src/action/platform_info.rs
```diff
@@ -0,0 +1,271 @@
...
+            node: self.node,
+            pep425tag: Some(
+                format!("{}_{}_{}", 
+                    self.system.unwrap_or(String::from("")), 
+                    self.release_name.unwrap_or(String::from("")), 
+                    self.architecture.unwrap_or(String::from(""))
+            )),
```
panhania, line 148: If any of these values is `None` you could end up with something like `foo__bar` or `_foo_`. Is this expected?

src/action/platform_info.rs
```diff
@@ -0,0 +1,271 @@
...
+            ..Default::default()
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
```
panhania, line 155: These tests mostly re-implement the implementation, which makes them quite useless (or even dangerous). In most cases, its of course impossible to write proper tests, so I think it is just better to assert that the values are non-empty.

src/action/mod.rs
```diff
@@ -15,6 +15,7 @@
 //! instance of the corresponding request type and send some (zero or more)
 //! instances of the corresponding response type.
 
+pub mod platform_info;
```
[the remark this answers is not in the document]
nikkita1267: This doesn't work, because my file is named `platform_info.rs` and if I change `platform_info` to `platform` I get compilation error. Or, maybe, I don't understand you correctly.

src/action/platform_info.rs
nikkita1267: Ok, thanks! Fixed.

src/action/platform_info.rs
nikkita1267: I need to clone, because `pep425tag` also try to move these fields.

src/action/platform_info.rs
nikkita1267: I decided to fill this field by analogy with the [Python code](https://github.com/google/grr/blob/master/grr/core/grr_response_core/lib/rdfvalues/client.py#L560-L566), but forgot that `None` converts to `'None'` in Python. Thanks!

src/action/platform_info.rs
nikkita1267: Ok, understand, thanks!

src/action/mod.rs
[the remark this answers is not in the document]
panhania: I mean, you should rename the module (both the file and this line) to be just `platform`.

src/action/platform_info.rs
panhania: Why? You are just stringifying them, borrowing them should be enough. You probably need to refactor it to a separate variable.

src/action/platform_info.rs
```diff
@@ -0,0 +1,236 @@
...
+    #[test]
+    fn test_system() {
+        let mut session = session::test::Fake::new();
+        assert!(handle(&mut session, ()).is_ok());
+
+        assert_eq!(session.reply_count(), 1);
+        let platform_info = &session.reply::<Response>(0);
+
+        assert!(platform_info.system.is_some());
+    }
```
panhania, line 169: For this test you could actually have platform-specific test cases for every supported platform (through `#[cfg_attr(..., ignore)]`) that actually test this field for some specific values (e.g. "Linux").

src/action/platform_info.rs
panhania: For example, for `system` you could have an enum with `Linux`, `Windows` etc. But this might not be ideal solution as it might break on some awkward platforms (that we don't really support, but who knows). On the other hand, passing through library-dependent values is not great either as we bind semantics of our protocol to some third-party internal details. In the end, I think it is fine to leave it as it is, I might look into that in the future.

src/action/platform_info.rs
```diff
@@ -0,0 +1,236 @@
...
+}
+
+impl From<Error> for session::Error {
+
+    fn from(error: Error) -> session::Error {
+        session::Error::action(error)
+    }
+}
+
+/// A Response type for `GetPlatformInfo` action
```
panhania, line 56: "Response" should not be uppercase here, there should be a dot at the end of the comment.

src/action/platform_info.rs
```diff
@@ -0,0 +1,236 @@
...
+#[derive(Default)]
+pub struct Response {
+    /// Client platform information
+    system: Option<String>,
```
panhania, line 60: This documentation item looks like something that applies to the whole struct, not only to the `system` field.

src/action/platform_info.rs
```diff
@@ -0,0 +1,236 @@
...
+    Ok(())
+}
+
+/// Handles requests for `GetPlatformInfo` action.
+pub fn handle<S: Session>(session:&mut S, _: ()) -> session::Result<()> {
+    let os_type = os_type().map_err(Error::CannotGetOSType)?;
+    match os_type.as_str() {
+        "Linux" => {
+            get_linux_response(session, os_type)?;
+        },
```
panhania, line 114: This looks like a good candidate for conditional compilation rather than some run-time behaviour that depends on what the third-party library decides to return.

PR #33, opened 2020-05-17
Add the `ListDirectory` action

### LINKED ISSUE #8: Implement the `ListDirectory` action
RRG should support the [`ListDirectory`][ls-action] action. It accepts [`ListDirRequest`][ls-request] as input and should respond with [`StatEntry`][ls-response].

[ls-action]: https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/client/grr_response_client/client_actions/standard.py#L160-L177
[ls-request]: https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/proto/grr_response_proto/jobs.proto#L682-L685
[ls-response]:
https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/proto/grr_response_proto/jobs.proto#L707-L802

M0rC1PanN:
This pull request implements the `ListDirectory` action, closes #8.


M0rC1PanN:
@googlebot I signed it!

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+}
+
+/// Converts enum type back to integer to pass to the proto
+fn get_int_path_type(pathspec: &PathSpec) -> Option<i32> {
+    match pathspec.pathtype {
+        PathType::OS => Some(0),
+        PathType::TSK => Some(1),
+        PathType::Registry => Some(2),
+        PathType::TMPFile => Some(3),
+        PathType::NTFS => Some(4),
```
panhania, line 236: Ditto.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+}
+
+/// Converts enum type back to integer to pass to the proto
+fn get_int_path_options(pathspec: &PathSpec) -> Option<i32> {
+    match pathspec.path_options {
+        Some(PathOption::CaseInsensitive) => Some(0),
+        Some(PathOption::CaseLiteral) => Some(1),
+        Some(PathOption::Recursive) => Some(2),
+        Some(PathOption::Regex) => Some(3),
+        _ => None
```
panhania, line 225: Ditto.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+        match ioctls::fs_ioc_getflags(file.as_raw_fd(), linux_flags_ptr) {
+            0 => Some(linux_flags),
+            _ => None,
+        };
+    }
+    Some(linux_flags)
+}
+
+/// Converts enum type back to integer to pass to the proto
+fn get_int_path_options(pathspec: &PathSpec) -> Option<i32> {
```
panhania, line 219: Is there a reason for this not being a method of `PathSpec`?

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+        Some(PathOption::CaseInsensitive) => Some(0),
+        Some(PathOption::CaseLiteral) => Some(1),
+        Some(PathOption::Recursive) => Some(2),
+        Some(PathOption::Regex) => Some(3),
+        _ => None
+    }
+}
+
+/// Converts enum type back to integer to pass to the proto
+fn get_int_path_type(pathspec: &PathSpec) -> Option<i32> {
```
panhania, line 230: Also here, why not a method?

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+                mount_point: None,
+                stream_name: None,
+                file_size_override: None,
+                inode: None,
+                is_virtualroot: None,
+                nested_path: None,
+                ntfs_id: None,
+                ntfs_type: None,
+                offset: None,
+                recursion_depth: None,
```
panhania, line 308: `..Default::default()`

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::action::Request;
+    use tempfile::tempdir;
+    use rand::thread_rng;
+    use rand::seq::SliceRandom;
+    use std::os::unix::fs::PermissionsExt;
```
panhania, line 324: Ditto.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+                              "Afile", "aafile", "afile",
+                              "bfile", "zfile", "*file", "(file)", "=^-^="];
+        let mut file_paths = vec![];
+        for file_name in &file_names {
+            file_paths.push(dir_path.join(file_name));
+        }
+        file_paths.shuffle(&mut thread_rng());
+        for file_path in &file_paths {
+            std::fs::File::create(file_path).unwrap();
+        }
```
panhania, line 433: Instead of this complicated logic, try to keep your tests simple. Here, I think it would be more readable to have explicit `std::fs::File::create` calls with explicit paths, instead of using `for`s and temporary arrays, e.g.:

```rust
std::fs::File::create(dir_path.join("foo")).unwrap();
std::fs::File::create(dir_path.join("bar")).unwrap();
```

I can see some value in having a different order each time you execute the tests, but I don't think the cost (unreadable test code) is worth it. It also introduces a problem, that (unlikely, but possible) your `shuffle` call with give you an already ordered input and the test will pass even if your action does not sort the results.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+                pathtype: PathType::OS,
+                path: PathBuf::from(&dir_path),
+            }
+        };
+        let mut session = session::test::Fake::new();
+        assert!(handle(&mut session, request).is_ok());
+        for i in 0..file_paths.len() {
+            assert_eq!(&session.reply::<Response>(i)
+                .pathspec.path, &file_paths[i]);
+        }
```
panhania, line 447: Similarly here, it is better to have repeated explicit assertions over using loops to reduce number of lines of code.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+        assert!(handle(&mut session, request).is_err());
+    }
+
+    #[test]
+    fn test_lexicographical_order() {
+        let dir = tempdir().unwrap();
+        let dir_path = dir.path();
+        let file_names = vec!["1numfile", "5numfile",
+                              "Afile", "aafile", "afile",
+                              "bfile", "zfile", "*file", "(file)", "=^-^="];
```
panhania, line 425: You can also include a filename with fancy unicode characters.

src/action/mod.rs
```diff
@@ -17,6 +17,8 @@
 
 pub mod metadata;
 pub mod startup;
+pub mod list_directory;
+// pub mod file_stat;
```
panhania, line 21: Commented-out code should not be checked into the repository.

src/action/mod.rs
```diff
@@ -93,6 +95,8 @@ where
     match action {
         "SendStartupInfo" => task.execute(self::startup::handle),
         "GetClientInfo" => task.execute(self::metadata::handle),
+        "ListDirectory" => task.execute(self::list_directory::handle),
+        // "StatFile" => task.execute(self::file_stat::handle),
```
panhania, line 99: Ditto.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+    st_size: u64,
+    st_atime: u64,
+    st_mtime: u64,
+    st_ctime: u64,
+    st_blocks: u32,
+    st_blksize: u32,
+    st_rdev: u32,
+    st_flags_linux: u32,
+    symlink: Option<String>,
+    pathspec: PathSpec,
```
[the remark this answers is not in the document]
M0rC1PanN: Well, I'm not sure, that u32 is more idiomatic to rust than u64 for example. I think changing `symlink` to  `Option<PathBuf>` would be reasonable here. But I don't know if we should change other fields.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+pub struct Request {
+    pathspec: PathSpec,
+}
+
+enum PathType {
+    OS,
+    TSK,
+    Registry,
+    TMPFile,
+    NTFS,
```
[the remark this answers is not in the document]
M0rC1PanN: Shall I remove PathType field at all and just fill response with OS type and throw an error otherwise?

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+enum PathOption {
+    CaseInsensitive,
+    CaseLiteral,
+    Regex,
+    Recursive,
+}
+
+struct PathSpec {
+    path_options: Option<PathOption>,
+    pathtype: PathType,
```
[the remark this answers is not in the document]
M0rC1PanN: Not yet, but `PathOption` will be needed if the code will be refactored to work on another OS, for instance.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+
+/// Converts integer from proto to human-readable enum type
+fn get_enum_path_options(option: &Option<i32>) -> Option<PathOption> {
+    match option {
+        Some(poption) => match poption {
+            0 => Some(PathOption::CaseInsensitive),
+            1 => Some(PathOption::CaseLiteral),
+            2 => Some(PathOption::Recursive),
+            3 => Some(PathOption::Regex),
+            _ => None
```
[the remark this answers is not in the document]
M0rC1PanN: It's conversion from protobuf. I don't see how not to rely on them.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+            st_ctime: umetadata.ctime() as u64,
+            st_blocks: umetadata.blocks() as u32,
+            st_blksize: umetadata.blksize() as u32,
+            st_rdev: umetadata.rdev() as u32,
+            st_flags_linux:
+            get_linux_flags(file_path).unwrap_or_default() as u32,
+            symlink: match umetadata.file_type().is_symlink() {
+                true => match fs::read_link(file_path) {
+                    Ok(file) => Some(file.to_string_lossy().to_string()),
+                    _ => None,
```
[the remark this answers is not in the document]
M0rC1PanN: I think, it's better not to throw error here to send some data about the file at least, if it's what you ment.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+use std::fs::{self, File};
+use std::path::PathBuf;
+use std::os::raw::c_long;
+use std::os::unix::fs::MetadataExt;
+use std::os::unix::io::AsRawFd;
+use std::fmt::{Display, Formatter};
+
+#[derive(Debug)]
+enum Error {
+    MissingFieldError(String),
```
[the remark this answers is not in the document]
M0rC1PanN: Fixed.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+    ReadPathError(std::io::Error),
```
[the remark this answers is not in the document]
M0rC1PanN: Fixed.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+impl Display for Error {
+    fn fmt(&self, fmt: &mut Formatter) -> std::fmt::Result {
+        use Error::*;
+
+        match *self {
+            MissingFieldError(ref field) => {
+                write!(fmt, "{} field should be filled", field)
+            }
+            ReadPathError(ref error) => {
+                write!(fmt, "Unable to read path: {}", error)
```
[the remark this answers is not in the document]
M0rC1PanN: Fixed.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+use std::fmt::{Display, Formatter};
+
+#[derive(Debug)]
+enum Error {
+    MissingFieldError(String),
+    ReadPathError(std::io::Error),
+}
+
+impl std::error::Error for Error {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
```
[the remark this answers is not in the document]
M0rC1PanN: Fixed.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+    let dir_path = &request.pathspec.path;
+    // In case if Result<DirEntry> matches the error branch, we won't give
+    // any info about that file, but other info isn't effected:
+    let (dir_entries, _): (Vec<_>, Vec<_>) = match dir_path.read_dir() {
+        Ok(dir_iter) => dir_iter,
+        Err(error) =>
+            return Err(session::Error::from(Error::ReadPathError(error))),
+    }.partition(Result::is_ok);
+    // Code won't panic because of unwrap(). We made the partition above
+    let mut paths: Vec<PathBuf> = dir_entries.into_iter().map(Result::unwrap)
```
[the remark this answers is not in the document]
M0rC1PanN: I found this solution in [Iterating over Results](https://doc.rust-lang.org/stable/rust-by-example/error/iter_result.html#collect-all-valid-values-and-failures-with-partition) article
And it seemed to be the clearest one for me.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+        .map(|entry| entry.path()).collect();
+    paths.sort();
+    for file_path in &paths {
+        let umetadata = match fs::symlink_metadata(file_path) {
```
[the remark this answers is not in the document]
M0rC1PanN: Fixed.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+
+pub fn handle<S: Session>(session: &mut S, request: Request)
+                          -> session::Result<()> {
+    let dir_path = &request.pathspec.path;
+    // In case if Result<DirEntry> matches the error branch, we won't give
+    // any info about that file, but other info isn't effected:
+    let (dir_entries, _): (Vec<_>, Vec<_>) = match dir_path.read_dir() {
+        Ok(dir_iter) => dir_iter,
+        Err(error) =>
+            return Err(session::Error::from(Error::ReadPathError(error))),
```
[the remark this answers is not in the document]
M0rC1PanN: Fixed.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+    // Code won't panic because of unwrap(). We made the partition above
+    let mut paths: Vec<PathBuf> = dir_entries.into_iter().map(Result::unwrap)
+        .map(|entry| entry.path()).collect();
+    paths.sort();
+    for file_path in &paths {
+        let umetadata = match fs::symlink_metadata(file_path) {
+            Ok(metadata) => metadata,
+            Err(error) =>
+                return Err(session::Error::from(Error::ReadPathError(error))),
+        };
```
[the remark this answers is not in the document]
M0rC1PanN: Fixed.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+            st_rdev: umetadata.rdev() as u32,
+            st_flags_linux:
+            get_linux_flags(file_path).unwrap_or_default() as u32,
+            symlink: match umetadata.file_type().is_symlink() {
+                true => match fs::read_link(file_path) {
+                    Ok(file) => Some(file.to_string_lossy().to_string()),
+                    _ => None,
+                }
+                false => None
+            },
```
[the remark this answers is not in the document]
M0rC1PanN: Fixed.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+
+fn get_path(path: &Option<String>) -> PathBuf {
+    match path {
+        Some(string_path) => {
+            if string_path.is_empty() {
+                PathBuf::from("/")
+            } else {
+                PathBuf::from(string_path)
+            }
+        }
```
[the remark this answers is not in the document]
M0rC1PanN: Fixed.

src/action/mod.rs
M0rC1PanN: Fixed.

src/action/mod.rs
M0rC1PanN: Fixed.

src/action/list_directory.rs
M0rC1PanN: Fixed.

src/action/list_directory.rs
M0rC1PanN: Fixed.

src/action/list_directory.rs
M0rC1PanN: Done.

src/action/list_directory.rs
```diff
@@ -251,20 +241,19 @@ impl super::Request for Request {
                         Some(path_type) => path_type,
                         None => return
                             Err(session::ParseError::malformed
-                                (Error::MissingFieldError
-                                    (String::from("path type"))))
+                                (session::MissingFieldError::new("path type")))
```
panhania, line 244: You can simplify this a lot by using error mapping. Take a look how it is done [here][1].

[1]: https://github.com/google/rrg/blob/8516f06c5398ccb137507fc9c0a43baf85a5fb53/src/session/demand.rs#L72-L77

src/action/list_directory.rs
```diff
@@ -420,18 +409,12 @@ mod tests {
...
-        for file_path in &file_paths {
-            std::fs::File::create(file_path).unwrap();
-        }
-        file_paths.sort();
+        std::fs::File::create(dir_path.join("Datei")).unwrap();
+        std::fs::File::create(dir_path.join("file")).unwrap();
+        std::fs::File::create(dir_path.join("αρχείο")).unwrap();
+        std::fs::File::create(dir_path.join("файл")).unwrap();
+        std::fs::File::create(dir_path.join("फ़ाइल")).unwrap();
+        std::fs::File::create(dir_path.join("❤ℝℝG❤")).unwrap();
```
panhania, line 417: 👍

src/action/list_directory.rs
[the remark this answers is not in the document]
panhania: For example, things like timestamps should use `SystemTime`. For the rest of things, at least in C, there are typedefs such as `dev_t`, `uid_t`, etc. I am not sure about Rust, though.

Also, the `st_` prefix looks very... C-like. Do we need it?

src/action/list_directory.rs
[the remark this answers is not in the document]
panhania: Unused code should generally not commited. Who knows, maybe whoever wants to work on this in the future, might take a different approach.

src/action/list_directory.rs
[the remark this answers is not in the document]
panhania: I think its the most reasonable approach for now.

src/action/list_directory.rs
```diff
@@ -0,0 +1,553 @@
...
+pub fn handle<S: Session>(session: &mut S, request: Request)
+                          -> session::Result<()> {
+    let dir_path = &request.pathspec.path;
+    // In case if Result<DirEntry> matches the error branch, we won't give
+    // any info about that file, but other info isn't effected:
+    let (dir_entries, _): (Vec<_>, Vec<_>) = dir_path.read_dir()
+        .map_err(Error::ReadPath)?.partition(Result::is_ok);
+    // Code won't panic because of unwrap(). We made the partition above
+    let mut paths: Vec<PathBuf> = dir_entries.into_iter().map(Result::unwrap)
+        .map(|entry| entry.path()).collect();
```
panhania, line 113: (I cannot reply to the previous thread on this issue)

I am pretty sure you can get rid of these unwraps if you use [`filter_map`].

[`filter_map`]: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter_map

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+    Registry,
+    TMPFile,
+    NTFS,
+}
+
+enum PathOption {
+    CaseInsensitive,
+    CaseLiteral,
+    Regex,
+    Recursive,
```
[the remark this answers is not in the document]
M0rC1PanN: Should I make `struct PathOption` with `bool` fields or exists a better approach?

src/action/list_directory.rs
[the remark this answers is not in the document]
M0rC1PanN: Done.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+}
+
+struct PathSpec {
+    path_options: Option<PathOption>,
+    pathtype: PathType,
+    path: PathBuf,
+}
+
+pub fn handle<S: Session>(session: &mut S, request: Request)
```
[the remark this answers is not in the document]
M0rC1PanN: Fixed.

src/action/list_directory.rs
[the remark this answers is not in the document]
M0rC1PanN: Fixed.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+        Ok(file) => file,
+        Err(_) => return None,
+    };
+    let mut linux_flags: c_long = 0;
+    let linux_flags_ptr: *mut c_long = &mut linux_flags;
+    unsafe {
+        match ioctls::fs_ioc_getflags(file.as_raw_fd(), linux_flags_ptr) {
+            0 => Some(linux_flags),
+            _ => None,
+        };
```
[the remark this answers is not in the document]
M0rC1PanN: Fixed.

src/action/list_directory.rs
```diff
@@ -108,45 +106,41 @@ pub fn handle<S: Session>(session: &mut S, request: Request)
...
                     Ok(file) => Some(file.to_string_lossy().to_string()),
                     _ => None,
                 }
-                false => None
+            } else {
+                None
             },
-            pathspec: PathSpec {
+            pathspec: PathSpec
+            {
```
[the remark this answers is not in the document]
M0rC1PanN: Fixed.

src/action/list_directory.rs
M0rC1PanN: Thanks, done.

src/action/list_directory.rs
M0rC1PanN: Yep, thank you, done.

src/action/list_directory.rs
M0rC1PanN: Done.

src/action/list_directory.rs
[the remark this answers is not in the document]
panhania: They are not used right now. Your action only works with literal paths (I suppose), so if during proto parsing you encounter something else, you can just fail.

However, even if there was a need to support these different path options, you should not rely on these magic numbers (which can change and suddenly your code will break without any compilation errors). Instead you should use associated enums (e.g. `rrg_proto::path_spec::Options`).

src/action/list_directory.rs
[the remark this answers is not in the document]
panhania: It depends. What options do you care about now?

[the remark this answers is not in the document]
M0rC1PanN: I'm not sure that any `OS` path(which we support now) is `CaseLiteral`. If it's so, I'll care only about that option and throw an error if another one comes from protobuf.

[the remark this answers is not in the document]
panhania: This is what I would do. I think its sane to assume that it is `CaseLiteral` (there is some magic with case correction in the file-finder flow but I think we can disregard this for now) and not have this logic at all.

src/action/list_directory.rs
```diff
@@ -0,0 +1,727 @@
...
+    gid: Option<u32>,
+    size: Option<u64>,
+    atime: Option<SystemTime>,
+    mtime: Option<SystemTime>,
+    ctime: Option<SystemTime>,
+    blocks: Option<u32>,
+    blksize: Option<u32>,
+    rdev: Option<u32>,
+    flags_linux: Option<u32>,
+    symlink: Option<String>,
```
panhania, line 127: Isn't symlink a path? What if it contains non-unicode characters?

src/action/list_directory.rs
```diff
@@ -0,0 +1,727 @@
...
+
+
+#[cfg(target_os = "linux")]
+fn get_status_change_time(metadata: &Metadata) -> Option<SystemTime> {
+    UNIX_EPOCH.checked_add(Duration::from_secs(metadata.ctime() as u64))
+}
+
+
+#[cfg(target_os = "linux")]
+fn fill_response(metadata: &Metadata, file_path: &PathBuf) -> Response {
```
panhania, line 215: Borrowing a `PathBuf` is not very idiomatic (the same as borrowing `String`). Prefer to use `Path` or `AsPath` trait instead.

src/action/list_directory.rs
```diff
@@ -0,0 +1,727 @@
...
+                          -> session::Result<()> {
+    let dir_path = &request.pathspec.path;
+    let mut paths: Vec<PathBuf> = dir_path.read_dir()
+        .map_err(Error::ReadPath)?.filter_map(|entry| entry.ok())
+        .map(|entry| entry.path()).collect();
+    paths.sort();
+
+    for file_path in &paths {
+        let metadata = fs::symlink_metadata(file_path)
+            .map_err(Error::ReadPath)?;
```
panhania, line 276: Doesn't it fit in one line?

src/action/list_directory.rs
```diff
@@ -0,0 +1,727 @@
...
+        },
+        _ => None,
+    }
+}
+
+fn get_path(path: &Option<String>) -> PathBuf {
+    match path {
+        Some(string_path) if !string_path.is_empty() => {
+            PathBuf::from(string_path)
+        }
```
panhania, line 301: I think it fits in one line, no need for braces.

src/action/mod.rs
```diff
@@ -100,14 +101,15 @@ where
...
+        "ListDirectory" => task.execute(self::list_directory::handle),
         "ListNetworkConnections" => task.execute(self::network::handle),
 
         #[cfg(target_family = "unix")]
         "EnumerateInterfaces" => task.execute(self::interfaces::handle),
 
         #[cfg(target_os = "linux")]
         "EnumerateFilesystems" => task.execute(self::filesystems::handle),
-
+
```
panhania, line 112: I don't think it should be here.

src/action/list_directory.rs
```diff
@@ -0,0 +1,727 @@
...
+use ioctls;
+
+#[cfg(target_os = "linux")]
+use std::os::raw::c_long;
+
+#[cfg(target_os = "linux")]
+use std::os::unix::fs::MetadataExt;
+
+#[cfg(target_os = "linux")]
+use std::os::unix::io::AsRawFd;
```
panhania, line 35: Prefer to have platform-specific imports in platform-specific functions that care about them. At the top level it just feels awkward.

src/action/list_directory.rs
```diff
@@ -0,0 +1,727 @@
...
+    UnsupportedValue(String, String),
+}
+
+impl std::error::Error for ParseError {
+
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        use ParseError::*;
+
+        match *self {
+            UnsupportedValue(ref _field, ref _value) => None,
```
panhania, line 84: You can just use a wildcard (`_`, no need for names).

src/action/list_directory.rs
```diff
@@ -0,0 +1,727 @@
...
+impl From<Error> for session::Error {
+
+    fn from(error: Error) -> session::Error {
+        session::Error::action(error)
+    }
+}
+
+#[derive(Debug)]
+enum ParseError {
+    UnsupportedValue(String, String),
```
panhania, line 75: Try to avoid tuples. Here, by just looking at the definition, I have no idea what these fields stand for.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+
+use crate::session::{self, Session};
+use rrg_proto::{ListDirRequest, StatEntry};
+
+use ioctls;
+use std::fs::{self, File};
+use std::path::PathBuf;
+use std::os::raw::c_long;
+use std::os::unix::fs::MetadataExt;
+use std::os::unix::io::AsRawFd;
```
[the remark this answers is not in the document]
M0rC1PanN: Fixed.

src/action/list_directory.rs
M0rC1PanN: Nope

src/action/list_directory.rs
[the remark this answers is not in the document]
M0rC1PanN: Done.

src/action/list_directory.rs
M0rC1PanN: It doesn't fit in 80 characters per line.

src/action/list_directory.rs
M0rC1PanN: Fixed.

src/action/list_directory.rs
M0rC1PanN: Replaced the tuple with a struct.

src/action/list_directory.rs
M0rC1PanN: Done.

src/action/mod.rs
M0rC1PanN: Fixed.

src/action/list_directory.rs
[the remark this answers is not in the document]
M0rC1PanN: Solved by removing all not needed options.

src/action/list_directory.rs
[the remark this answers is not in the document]
M0rC1PanN: Fixed.

src/action/list_directory.rs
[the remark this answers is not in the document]
M0rC1PanN: Fixed.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+
+/// Converts integer from proto to human-readable enum type
+fn get_enum_path_type(option: &Option<i32>) -> Option<PathType> {
+    match option {
+        Some(ptype) => match ptype {
+            0 => Some(PathType::OS),
+            1 => Some(PathType::TSK),
+            2 => Some(PathType::Registry),
+            3 => Some(PathType::TMPFile),
+            4 => Some(PathType::NTFS),
```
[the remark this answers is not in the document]
M0rC1PanN: Fixed.

src/action/list_directory.rs
M0rC1PanN: Fixed.

src/action/list_directory.rs
M0rC1PanN: I removed all such functions.

src/action/list_directory.rs
M0rC1PanN: I've removed all such functions.

src/action/list_directory.rs
M0rC1PanN: Fixed.

src/action/list_directory.rs
M0rC1PanN: Fixed.

src/action/list_directory.rs
```diff
@@ -0,0 +1,533 @@
...
+    st_blksize: u32,
+    st_rdev: u32,
+    st_flags_linux: u32,
+    symlink: Option<String>,
+    pathspec: PathSpec,
+}
+
+/// A request type for the list directory action.
+pub struct Request {
+    pathspec: PathSpec,
```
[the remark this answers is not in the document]
M0rC1PanN: Fixed, I suppose.

src/action/list_directory.rs
M0rC1PanN: I've refactored this field. It's `Option<PathBuf>` now.

src/action/list_directory.rs
```diff
@@ -382,14 +354,19 @@ impl super::Response for Response {
             st_rdev: self.rdev,
             st_flags_osx: None,
             st_flags_linux: self.flags_linux,
-            symlink: self.symlink,
+            symlink:
+            self.symlink
+                .map_or(None,
+                        |symlink| Some(symlink.to_string_lossy().to_string())),
```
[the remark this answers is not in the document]
M0rC1PanN: I'm not sure that I get it correctly. Do you mean to have `Some("")` instead of `None`?

src/action/list_directory.rs
```diff
@@ -70,9 +56,24 @@ impl From<Error> for session::Error {
...
+
+impl std::fmt::Debug for UnsupportedValueMessage {
+
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        f.debug_struct("UnsupportedValueMessage")
+            .field("field", &self.field)
+            .field("value", &self.value)
+            .finish()
+    }
+}
```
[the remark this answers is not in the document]
M0rC1PanN: Fixed.

src/action/list_directory.rs
```diff
@@ -70,9 +56,24 @@ impl From<Error> for session::Error {
     }
 }
 
+struct UnsupportedValueMessage {
```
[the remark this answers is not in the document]
M0rC1PanN: Done.

src/action/list_directory.rs
```diff
@@ -340,16 +302,26 @@ impl super::Request for Request {
...
             .ok_or(missing("path type"))?;
-        if path_type != 0 {
+        if path_type != PathType::Os as i32 {
             return Err(session::ParseError::malformed
                 (ParseError::UnsupportedValue
-                    (String::from("path type"), path_type.to_string())));
+                    (UnsupportedValueMessage {
+                        field: String::from("path type"),
+                        value: path_type.to_string(),
+                    })));
```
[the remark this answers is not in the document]
M0rC1PanN: Done.

src/action/list_directory.rs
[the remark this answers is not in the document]
M0rC1PanN: Done, thx.

src/action/list_directory.rs
```diff
@@ -382,14 +354,19 @@ impl super::Response for Response {
...
+            symlink:
+            self.symlink
+                .map_or(None,
+                        |symlink| Some(symlink.to_string_lossy().to_string())),
             registry_type: None,
             resident: None,
             pathspec: Some(rrg_proto::PathSpec {
-                path_options: get_int_path_options(&self.pathspec),
+                // represents CaseLiteral path option (other options are not
+                // supported)
```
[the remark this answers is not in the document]
M0rC1PanN: Fixed.

src/action/list_directory.rs
```diff
@@ -436,21 +413,42 @@ mod tests {
     #[test]
     fn test_empty_path_options() {
         let request: Result<super::Request, _> = Request::from_proto
-            (fill_proto_request(None, Some(PathType::Os as i32), Some(String::from("/"))));
+            (fill_proto_request(None, Some(PathType::Os as i32),
```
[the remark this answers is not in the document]
M0rC1PanN: Fixed.

src/action/list_directory.rs
```diff
@@ -56,24 +56,23 @@ impl From<Error> for session::Error {
...
-            .field("field", &self.field)
-            .field("value", &self.value)
-            .finish()
-    }
+fn unsupported_value(field: String, value: String) -> session::ParseError {
+    session::ParseError::malformed(ParseError::UnsupportedValue(
+        UnsupportedValueError {
+            field,
+            value,
+        }))
```
panhania, line 70: This should be a method of `ParseError`. Also, it is missing documentation.

src/action/list_directory.rs
```diff
@@ -356,15 +347,14 @@ impl super::Response for Response {
...
             registry_type: None,
             resident: None,
             pathspec: Some(rrg_proto::PathSpec {
-                // represents CaseLiteral path option (other options are not
-                // supported)
+                // Represents `CaseLiteral` path option (other options are not
+                // supported).
                 path_options: Some(Options::CaseLiteral as i32),
-                // represents OS path type (other types are not supported)
+                // Represents `OS` path type (other types are not supported).
```
panhania, line 357: "OS" is actually not an identifier, it's just an abbreviation of "operating system", so should not be backticked.

src/action/list_directory.rs
```diff
@@ -0,0 +1,689 @@
...
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::action::Request;
+    use tempfile::tempdir;
+
+    #[cfg(target_os = "linux")]
+    use std::os::unix::fs::MetadataExt;
+
+    /// Fills ListDirRequest with provided fields
```
panhania, line 378: `ListDirRequest` should be backticked. Also, a missing dot at the end.

src/action/list_directory.rs
panhania: It will if you rename `string_path` to `path` :)

src/action/list_directory.rs
panhania: :(

src/action/list_directory.rs
```diff
@@ -0,0 +1,689 @@
...
+        }
+    }
+}
+
+/// A request type for the list directory action.
+pub struct Request {
+    path: PathBuf,
+}
+
+fn get_accesses_time(metadata: &Metadata) -> Option<SystemTime> {
```
panhania, line 162: `get_access_time`, not `get_accesses_time`

src/action/list_directory.rs
```diff
@@ -0,0 +1,689 @@
...
+    mode: Option<u64>,
+    ino: Option<u32>,
+    dev: Option<u32>,
+    nlink: Option<u32>,
+    uid: Option<u32>,
+    gid: Option<u32>,
+    size: Option<u64>,
+    atime: Option<SystemTime>,
+    mtime: Option<SystemTime>,
+    ctime: Option<SystemTime>,
```
panhania, line 122: Some of these fields are always `Some`. No need to make them `Optional` then.

src/action/list_directory.rs
```diff
@@ -0,0 +1,689 @@
...
+use log::warn;
+use std::time::{SystemTime, UNIX_EPOCH};
+
+#[cfg(target_os = "linux")]
+use std::os::raw::c_long;
+
+#[derive(Debug)]
+enum Error {
+    ReadPath(std::io::Error),
+}
```
panhania, line 26: Missing documentation.

src/action/list_directory.rs
```diff
@@ -0,0 +1,689 @@
...
+        gid: Some(metadata.gid() as u32),
+        size: Some(metadata.size()),
+        atime: get_accesses_time(&metadata),
+        mtime: get_modification_time(&metadata),
+        ctime: get_status_change_time(&metadata),
+        blocks: Some(metadata.blocks() as u32),
+        blksize: Some(metadata.blksize() as u32),
+        rdev: Some(metadata.rdev() as u32),
+        flags_linux:
+        Some(get_linux_flags(file_path).unwrap_or_default() as u32),
```
panhania, line 220: Why `Some` in case you fail to obtain the flags?

src/action/list_directory.rs
```diff
@@ -0,0 +1,689 @@
...
+            match fs::read_link(file_path) {
+                Ok(file) => Some(file),
+                Err(error) => {
+                    warn!("unable to read symlink: {}", error);
+                    None
+                }
+            }
+        } else {
+            None
+        },
```
panhania, line 231: This has some non-trivial logic and does I/O, it should not be inside struct initialization.

src/action/list_directory.rs
```diff
@@ -0,0 +1,689 @@
...
+        Err(err) => {
+            warn!("unable to get last access time: {}", err);
+            None
+        }
+    }
+}
+
+fn get_modification_time(metadata: &Metadata) -> Option<SystemTime> {
+    match metadata.modified() {
+        Ok(atime) => Some(atime),
```
panhania, line 174: Why `atime` for modification time?

src/action/list_directory.rs
```diff
@@ -0,0 +1,689 @@
...
+        Err(err) => {
+            warn!("unable to get last modification time: {}", err);
+            None
+        }
+    }
+}
+
+fn get_creation_time(metadata: &Metadata) -> Option<SystemTime> {
+    match metadata.created() {
+        Ok(atime) => Some(atime),
```
panhania, line 184: Ditto.

src/action/list_directory.rs
```diff
@@ -0,0 +1,689 @@
...
+        Ok(Request {
+            path: get_path(&pathspec.path),
+        })
+    }
+}
+
+fn get_time_since_unix_epoch(sys_time: &Option<SystemTime>) -> Option<u64> {
+    return sys_time.map_or(None, |time| time.duration_since(UNIX_EPOCH)
+        .map_or(None, |dur| Some(dur.as_secs())));
+}
```
panhania, line 323: I think you can use [this](https://github.com/google/rrg/blob/master/proto/src/lib.rs#L218-L221).

src/action/list_directory.rs
M0rC1PanN: Only `size`, I think.

src/action/list_directory.rs
M0rC1PanN: Fixed.

src/action/list_directory.rs
M0rC1PanN: It can be treated as abbreviation or one of enum options in protobuf(same as `CaseLiteral`), I suppose. Nevertheless, fixed.

src/action/list_directory.rs
M0rC1PanN: Fixed.

src/action/list_directory.rs
M0rC1PanN: Fixed.

src/action/list_directory.rs
M0rC1PanN: I've put this code into a separate function.

src/action/list_directory.rs
M0rC1PanN: Fixed.

src/action/list_directory.rs
M0rC1PanN: Fixed.

src/action/list_directory.rs
M0rC1PanN: Done, thx.

src/action/list_directory.rs
```diff
@@ -56,20 +55,28 @@ impl From<Error> for session::Error {
     }
 }
 
+/// Contains information about an `UnsupportedValue` error.
```
panhania, line 58: I mean, this doc string is not very useful. It still doesn't explain what an `UnsupportedValue` error is and in what circumstances can occur.

src/action/list_directory.rs
```diff
@@ -169,81 +177,95 @@ fn get_accesses_time(metadata: &Metadata) -> Option<SystemTime> {
...
             None
         }
     }
 }
 
+/// Returns the creation time of provided `Metadata`.
 fn get_creation_time(metadata: &Metadata) -> Option<SystemTime> {
     match metadata.created() {
-        Ok(atime) => Some(atime),
+        Ok(creation_time) => Some(creation_time),
```
panhania, line 194: This is inconsistent (other functions use `atime` and `mtime`), why this one does not follow the pattern?

src/action/list_directory.rs
```diff
@@ -169,81 +177,95 @@ fn get_accesses_time(metadata: &Metadata) -> Option<SystemTime> {
...
+    use std::time::{Duration, UNIX_EPOCH};
     use std::os::unix::fs::MetadataExt;
 
     UNIX_EPOCH.checked_add(Duration::from_secs(metadata.ctime() as u64))
 }
 
+/// Reads a symbolic link, returning the path to the file that the link points to.
+#[cfg(target_os = "linux")]
+fn get_symlink(metadata: &Metadata, file_path: &Path) -> Option<PathBuf> {
+    if metadata.file_type().is_symlink() {
```
panhania, line 214: Prefer to use early return to avoid nesting.

src/action/list_directory.rs
```diff
@@ -0,0 +1,730 @@
...
+
+    /// Constructs an `UnsupportedValue` from provided `field` and
+    /// `value`, then  converts it to `ParseError`.
+    fn unsupported_value(field: String, value: String) -> session::ParseError {
+        session::ParseError::malformed(ParseError::UnsupportedValue(
+            UnsupportedValueError {
+                field,
+                value,
+            }))
+    }
```
panhania, line 75: This should return `ParseError` (not `session::ParseError`), right now it looks a bit awkward: you have a method on type `ParseError` that returns completely unrelated type. Returning `ParseError` combined with proper `.into()` call (see a comment below) should make the error handle simple enough.

src/action/list_directory.rs
```diff
@@ -0,0 +1,730 @@
...
+        }
+    }
+}
+
+impl From<ParseError> for session::Error {
+
+    fn from(error: ParseError) -> session::Error {
+        session::Error::action(error)
+    }
+}
```
panhania, line 115: This should be implemented for `session::ParseError`, not action execution-specific `session::Error`.

src/action/mod.rs
```diff
@@ -23,6 +23,7 @@ pub mod interfaces;
 
 pub mod metadata;
 pub mod startup;
+pub mod list_directory;
```
panhania, line 26: I could have sworn that I left a comment about renaming the module, but I cannot find it. So far I tried to make all action names one-word identifiers, so lets try to keep this tradition for now. Something like `listdir` would work fine, I think.

src/action/list_directory.rs
```diff
@@ -0,0 +1,730 @@
...
+            symlink:
+            self.symlink
+                .map(|symlink| symlink.to_string_lossy().to_string()),
+            registry_type: None,
+            resident: None,
+            pathspec: Some(rrg_proto::PathSpec {
+                // Represents `CaseLiteral` path option (other options are not
+                // supported).
+                path_options: Some(Options::CaseLiteral as i32),
+                // Represents OS path type (other types are not supported).
```
panhania, line 388: I cannot reply to the previous thread (I can make only mark it as resolved), so I will just reply here: good point about "OS" also meaning an enum-specific value. In that case, I am not sure what is better after all.

src/action/list_directory.rs
M0rC1PanN: Ok, a reason for that is probably my misunderstanding of this `struct` purpose. I think, it's just a wrap on a `tuple` for `UnsupportedValue` fields. That's why it was called `UnsupportedValueMessage` at first.

src/action/list_directory.rs
M0rC1PanN: Fixed.

src/action/list_directory.rs
M0rC1PanN: Fixed.

src/action/list_directory.rs
M0rC1PanN: Done.

src/action/list_directory.rs
M0rC1PanN: Fixed.

src/action/mod.rs
M0rC1PanN: Done.

src/action/listdir.rs
```diff
@@ -323,15 +320,16 @@ impl super::Request for Request {
         let path_type = pathspec.pathtype
             .ok_or(missing("path type"))?;
         if path_type != PathType::Os as i32 {
-            return Err(ParseError::unsupported_value(String::from("path type"),
-                                                     path_type.to_string()));
+            return Err(session::ParseError::from(
+                ParseError::unsupported_value(String::from("path type"),
+                                              path_type.to_string())));
```
panhania, line 325: You can simplify `session::ParseError::from(something)` with `something.into()`.

src/action/listdir.rs
```diff
@@ -323,15 +320,16 @@ impl super::Request for Request {
...
         }
         let path_option = pathspec.path_options
             .unwrap_or(Options::CaseLiteral as i32);
         if path_option != Options::CaseLiteral as i32 {
-            return Err(ParseError::unsupported_value(
-                String::from("path option"),
-                path_option.to_string()));
+            return Err(session::ParseError::from(
+                ParseError::unsupported_value(String::from("path option"),
+                                              path_option.to_string())));
```
panhania, line 332: Ditto.

src/action/listdir.rs
```diff
@@ -66,12 +66,11 @@ impl ParseError {
...
-        session::ParseError::malformed(ParseError::UnsupportedValue(
-            UnsupportedValueError {
-                field,
-                value,
-            }))
+    fn unsupported_value(field: String, value: String) -> ParseError {
+        ParseError::UnsupportedValue(UnsupportedValueError {
+            field,
+            value,
+        })
```
panhania, line 73: You can make this function more general (which will simplify the code below), if instead of taking `String`, you take objects of type `Into<String>`. This way strings can still be passed into your function (because `Into<T>` is implemented for every `T` itself), but also things like string literals. The signature would like like:

```
fn unuspported_value<F, V>(field: F, value: V) -> ParseError
where
    F: Into<String>,
    V: Into<String>,
{ ... }
```

src/action/listdir.rs
panhania: Moreover, once you apply the suggestion above about generalization of `unsupported_value`, you can simplify the `String::from("path type")` with just `"path_type"`.

src/action/list_directory.rs
M0rC1PanN: I'm not sure, this issue is fixed.
But from my point of view this documentation string will make a person who reads it to find `UnsupportedValue` which is documented.

src/action/listdir.rs
```diff
@@ -0,0 +1,732 @@
...
+}
+
+impl ParseError {
+
+    /// Constructs an `UnsupportedValue` from provided `field` and
+    /// `value`, then  converts it to `ParseError`.
+    fn unsupported_value<F, V>(field: F, value: V) -> ParseError
+        where
+            F: Into<String>,
+            V: Into<String>,
```
panhania, line 72: `where` should be indented on the same level as `fn`.

src/action/listdir.rs
```diff
@@ -0,0 +1,732 @@
...
+/// Constructs `PathBuf` from `String`. If provided string is empty constructs
+/// `PathBuf` from "/".
+fn get_path(path: &Option<String>) -> PathBuf {
+    match path {
+        Some(path) if !path.is_empty() => PathBuf::from(path),
+        _ => PathBuf::from("/"),
+    }
+}
+
+/// Fills st_linux_flags field.
```
panhania, line 298: The `st_linux_flags` is an identifier and should use backticks.

src/action/listdir.rs
```diff
@@ -0,0 +1,732 @@
...
+        Ok(Request {
+            path: get_path(&pathspec.path),
+        })
+    }
+}
+
+/// Converts idiomatic `SystemTime` to `u64` for the protocol buffer.
+fn get_time_since_unix_epoch(sys_time: &Option<SystemTime>) -> Option<u64> {
+    match sys_time {
+        Some(time) => match micros(time.clone()) {
```
panhania, line 347: You can use `Some(ref time)` pattern, this way you don't need to use `.clone()`.

src/action/listdir.rs
```diff
@@ -0,0 +1,732 @@
...
+
+    Ok(())
+}
+
+/// Constructs `PathBuf` from `String`. If provided string is empty constructs
+/// `PathBuf` from "/".
+fn get_path(path: &Option<String>) -> PathBuf {
+    match path {
+        Some(path) if !path.is_empty() => PathBuf::from(path),
+        _ => PathBuf::from("/"),
```
panhania, line 294: Prefer to use explicit `None` over `_`. In general, it is better to have explicit patterns for enum types, as they will make the compiler tell you that you should revisit it when you add new variants or change types. Using wildcards will usually make it silently compile, but possibly with incorrect behaviour.

src/action/listdir.rs
panhania: Actually, `SystemTime` implements `Copy`, so you can just dereference it (`*time`, rather than `time.clone()`).

src/action/listdir.rs
M0rC1PanN: The `pattern `&Some(_)` not covered` error occurs in that case (`None` instead of `_`). It happens because of `if` condition: `_` here covers empty `path` and `None` variants.

panhania: Oh, sorry, you are right, of course. I didn't notice the guard clause.

PR #34, opened 2020-05-29
Add the `GetFileStat` action

### LINKED ISSUE #2: Implement the `GetFileStat` action
RRG should support the [`GetFileStat`][stat-action] action. It accepts [`GetFileStatRequest`][stat-request] as input and should respond with [`StatEntry`][stat-response].

[stat-action]: https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/client/grr_response_client/client_actions/standard.py#L186-L201
[stat-request]: https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/proto/grr_response_proto/jobs.proto#L687-L691
[stat-response]:
https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/proto/grr_response_proto/jobs.proto#L707-L802

RainMustFall:
This pull request implements the GetFileStat action, closes #2

### DIFF
--- a/src/action/file_stat.rs
+++ b/src/action/file_stat.rs
@@ -120,7 +120,6 @@ pub fn handle<S: Session>(session: &mut S, request: Request) -> session::Result<
 
 #[cfg(target_family = "unix")]
 fn get_ext_attrs(path: &Path) -> Vec<ExtAttr> {
-    use xattr;
     let xattrs = xattr::list(path).unwrap();
 
     let mut result = vec![];
--- a/src/action/file_stat.rs
+++ b/src/action/file_stat.rs
@@ -17,6 +17,8 @@ use std::os::raw::c_long;
 use std::os::unix::fs::MetadataExt;
 use std::os::unix::io::AsRawFd;
 use xattr;
+use log::warn;
+use std::time::{SystemTime, UNIX_EPOCH};
 
 impl From<std::io::Error> for Error {
     fn from(e: std::io::Error) -> Error {
@@ -25,20 +27,20 @@ impl From<std::io::Error> for Error {
 }
 
 pub struct Response {
-    mode: u64,
-    inode: u32,
-    device: u32,
-    hard_links: u32,
-    uid: u32,
-    gid: u32,
-    size: u64,
-    access_time: u64,
-    modification_time: u64,
-    status_change_time: u64,
-    blocks_number: u32,
-    block_size: u32,
-    represented_device: u32,
-    flags_linux: u32,
+    mode: Option<u64>,
+    inode: Option<u32>,
+    device: Option<u32>,
+    hard_links: Option<u32>,
+    uid: Option<u32>,
+    gid: Option<u32>,
+    size: Option<u64>,
+    access_time: Option<SystemTime>,
+    modification_time: Option<SystemTime>,
+    status_change_time: Option<SystemTime>,
+    blocks_number: Option<u32>,
+    block_size: Option<u32>,
+    represented_device: Option<u32>,
+    flags_linux: Option<u32>,
     symlink: Option<String>,
     pathspec: PathSpec,
     extended_attributes: Vec<rrg_proto::stat_entry::ExtAttr>,
@@ -114,26 +116,41 @@ fn get_ext_attrs(path: &PathBuf) -> Vec<rrg_proto::stat_entry::ExtAttr> {
     result
 }
 
+fn get_time_option<E: std::fmt::Display>(time: Result<SystemTime, E>) -> Option<SystemTime> {
+    match time {
+        Ok(time_value) => Some(time_value),
+        Err(err) => {
+            warn!("Unable to get time value: {}", err);
+            None
+        }
+    }
+}
+
+fn get_time_since_unix_epoch(sys_time: &Option<SystemTime>) -> Option<u64> {
+    return sys_time.map_or(None, |time| time.duration_since(UNIX_EPOCH)
+        .map_or(None, |dur| Some(dur.as_secs())));
+}
+
 fn form_response(original_path: &PathBuf, destination: &PathBuf)
                  -> Result<Response, std::io::Error> {
     let metadata = fs::symlink_metadata(destination)?;
     let original_metadata = fs::symlink_metadata(original_path)?;
 
     Ok(Response {
-        mode: metadata.mode().into(),
-        inode: metadata.ino() as u32,
-        device: metadata.dev() as u32,
-        hard_links: metadata.nlink() as u32,
-        uid: metadata.uid() as u32,
-        gid: metadata.gid() as u32,
-        size: metadata.size(),
-        access_time: metadata.atime() as u64,
-        modification_time: metadata.mtime() as u64,
-        status_change_time: metadata.ctime() as u64,
-        blocks_number: metadata.blocks() as u32,
-        block_size: metadata.blksize() as u32,
-        represented_device: metadata.rdev() as u32,
-        flags_linux: get_linux_flags(destination).unwrap_or_default() as u32,
+        mode: Some(metadata.mode() as u64),
+        inode: Some(metadata.ino() as u32),
+        device: Some(metadata.dev() as u32),
+        hard_links: Some(metadata.nlink() as u32),
+        uid: Some(metadata.uid() as u32),
+        gid: Some(metadata.gid() as u32),
+        size: Some(metadata.size() as u64),
+        access_time: get_time_option(metadata.accessed()),
+        modification_time: get_time_option(metadata.modified()),
+        status_change_time: get_time_option(metadata.created()),
+        blocks_number: Some(metadata.blocks() as u32),
+        block_size: Some(metadata.blksize() as u32),
+        represented_device: Some(metadata.rdev() as u32),
+        flags_linux: get_linux_flags(destination),
 
         symlink: match original_metadata.file_type().is_symlink() {
             true => Some(fs::read_link(original_path).
@@ -188,7 +205,7 @@ fn collapse_pathspec(pathspec: PathSpec) -> PathBuf {
                     Some(nested_path_box) => path.join(recursive_collapse(*nested_path_box)),
                     None => path,
                 }
-            },
+            }
             None => PathBuf::default()
         }
     }
@@ -207,7 +224,7 @@ fn get_path(path: &Option<String>) -> Option<PathBuf> {
     }
 }
 
-fn get_linux_flags(path: &PathBuf) -> Option<c_long> {
+fn get_linux_flags(path: &PathBuf) -> Option<u32> {
     let file = match File::open(path) {
         Ok(file) => file,
         Err(_) => return None,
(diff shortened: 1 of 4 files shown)


RainMustFall:
@googlebot I signed it!

src/action/file_stat.rs
```diff
@@ -0,0 +1,532 @@
...
+    access_time: Option<SystemTime>,
+    modification_time: Option<SystemTime>,
+    status_change_time: Option<SystemTime>,
+    blocks_number: Option<u32>,
+    block_size: Option<u32>,
+    represented_device: Option<u32>,
+    flags_linux: Option<u32>,
+    symlink: Option<PathBuf>,
+    pathspec: PathSpec,
+    extended_attributes: Vec<ExtAttr>,
```
panhania, line 44: By using a proto field here, your `Response` contains a state that is not really possible: an extended attribute with missing fields. Better to define your own structure where impossible cases are ruled-out on the type level.

This is why in general you should not use any types from the `rrg-proto` crate directly and convert from and into it in appropriate methods.

src/action/file_stat.rs
```diff
@@ -0,0 +1,532 @@
...
+}
+
+#[derive(Debug)]
+pub struct Request {
+    pathspec: Option<PathSpec>,
+    collect_ext_attrs: Option<bool>,
+    follow_symlink: Option<bool>,
```
panhania, line 51: Why are these optional? Can your action work if no path is specified? Also, having `Option<bool>` in general is very dubious.

src/action/file_stat.rs
```diff
@@ -0,0 +1,532 @@
...
+            warn!("Unable to get time value: {}", err);
+            None
+        }
+    }
+}
+
+fn get_time_since_unix_epoch(sys_time: &Option<SystemTime>) -> Option<u64> {
+    return sys_time.map_or(None, |time| time.duration_since(UNIX_EPOCH)
+        .map_or(None, |dur| Some(dur.as_secs())));
+}
```
panhania, line 153: Isn't [this](https://github.com/google/rrg/blob/master/proto/src/lib.rs#L218-L221) what you need?

src/action/mod.rs
```diff
@@ -18,6 +18,7 @@
 pub mod metadata;
 pub mod startup;
 pub mod network;
+pub mod file_stat;
```
panhania, line 21: Just `stat`.

src/action/file_stat.rs
```diff
@@ -0,0 +1,532 @@
...
+#[cfg(target_os = "linux")]
+fn get_status_change_time(metadata: &Metadata) -> Option<SystemTime> {
+    use std::time::Duration;
+    use std::os::unix::fs::MetadataExt;
+
+    UNIX_EPOCH.checked_add(Duration::from_secs(metadata.ctime() as u64))
+}
+
+#[cfg(target_os = "linux")]
+fn form_response(original_path: &PathBuf, destination: &PathBuf)
```
panhania, line 164: Passing `PathBuf` as a reference is an anti-pattern, use `Path` or `AsRef<Path>` instead (take hint from [the standard library](https://doc.rust-lang.org/std/fs/fn.metadata.html)).

src/action/file_stat.rs
```diff
@@ -0,0 +1,532 @@
...
+    hard_links: Option<u32>,
+    uid: Option<u32>,
+    gid: Option<u32>,
+    size: Option<u64>,
+    access_time: Option<SystemTime>,
+    modification_time: Option<SystemTime>,
+    status_change_time: Option<SystemTime>,
+    blocks_number: Option<u32>,
+    block_size: Option<u32>,
+    represented_device: Option<u32>,
```
panhania, line 40: Why are these optional? It should be easy to obtain these from `Metadata` on any system. You shouldn't redefine what `Metadata` has anyway in the first place.

src/action/file_stat.rs
```diff
@@ -0,0 +1,532 @@
...
+use std::time::{SystemTime, UNIX_EPOCH};
+
+use log::warn;
+use rrg_proto::{GetFileStatRequest, path_spec::Options, path_spec::PathType,
+                stat_entry::ExtAttr, StatEntry};
+
+use crate::session::{self, Error, Session};
+
+impl From<std::io::Error> for Error {
+    fn from(e: std::io::Error) -> Error {
```
panhania, line 21: I tend to have a blank newline between `impl`, `struct` and alike and the first item.

src/action/file_stat.rs
```diff
@@ -0,0 +1,532 @@
...
+#[derive(Debug)]
+pub struct Request {
+    pathspec: Option<PathSpec>,
+    collect_ext_attrs: Option<bool>,
+    follow_symlink: Option<bool>,
+}
+
+#[derive(Debug)]
+struct PathSpec {
+    nested_path: Option<Box<PathSpec>>,
```
panhania, line 56: `nested_paths` are used only with combination with TSK, no need for this for now.

src/action/file_stat.rs
```diff
@@ -0,0 +1,532 @@
...
+}
+
+#[cfg(target_os = "linux")]
+fn get_linux_flags(path: &Path) -> Option<u32> {
+    use std::os::raw::c_long;
+    use std::os::unix::io::AsRawFd;
+
+    let file = match File::open(path) {
+        Ok(file) => file,
+        Err(_) => return None,
```
panhania, line 241: I don't think we should ignore such errors, it warrants a log entry.

src/action/file_stat.rs
```diff
@@ -0,0 +1,532 @@
...
+    }
+}
+
+pub fn handle<S: Session>(session: &mut S, request: Request) -> session::Result<()> {
+    let original_path = collapse_pathspec(request.pathspec.unwrap());
+
+    let follow_symlink = match request.follow_symlink {
+        Some(s) => s,
+        None => false,
+    };
```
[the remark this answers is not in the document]
RainMustFall: Fixed.

src/action/file_stat.rs
```diff
@@ -0,0 +1,532 @@
...
+
+    let collect_ext_attrs = match request.collect_ext_attrs {
+        Some(s) => s,
+        None => false,
+    };
```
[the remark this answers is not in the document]
RainMustFall: Fixed.

src/action/file_stat.rs
```diff
@@ -0,0 +1,532 @@
...
+
+#[cfg(target_family = "unix")]
+fn get_ext_attrs(path: &Path) -> Vec<ExtAttr> {
+    let xattrs = xattr::list(path).unwrap();
+
+    let mut result = vec![];
+    for attr in xattrs {
+        result.push(ExtAttr {
+            name: Some(attr.to_str().unwrap().as_bytes().to_vec()),
+            value: xattr::get(path, attr).unwrap(),
```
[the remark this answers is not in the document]
RainMustFall: Well, I rewrote the first line so the extra `unwrap` call was avoided. However, I didn't find any guarantees about `xattr::get` in the documentation, so a check was added.

src/action/file_stat.rs
RainMustFall: Fixed.

src/action/file_stat.rs
RainMustFall: Fixed.

src/action/file_stat.rs
RainMustFall: Yes, fixed, thanks.

src/action/mod.rs
RainMustFall: Fixed.

src/action/file_stat.rs
RainMustFall: Fixed.

src/action/file_stat.rs
RainMustFall: Fixed.

src/action/file_stat.rs
RainMustFall: Fixed.

src/action/file_stat.rs
RainMustFall: Well, the main reason I had to add this is that without `nested_paths` support I couldn't test this action through the UI. However, I can always remove this feature if needed.

src/action/file_stat.rs
RainMustFall: Fixed.

src/action/stat.rs
```diff
@@ -46,18 +46,28 @@ pub struct Response {
...
+    name: Vec<u8>,
+    value: Vec<u8>,
 }
 
 #[derive(Debug)]
 struct PathSpec {
+
     nested_path: Option<Box<PathSpec>>,
     path: Option<PathBuf>,
 }
```
panhania: Since you collapse the path as the first thing in you `handle` method, why not just collapse it as part of the parsing of the proto? Then you don't need this type at all and just use idiomatic `PathBuf`s everywhere.

src/action/stat.rs
```diff
@@ -120,14 +106,19 @@ pub fn handle<S: Session>(session: &mut S, request: Request) -> session::Result<
...
-            name: Some(attr.to_str().unwrap().as_bytes().to_vec()),
-            value: xattr::get(path, attr).unwrap(),
-        });
+        match xattr::get(path, &attr) {
+            Ok(attr_value) => result.push(ExtAttr {
+                name: attr.into_vec(),
+                value: attr_value.unwrap_or_default(),
+            }),
+
+            Err(_) => ()
```
panhania, line 120: I would at least log this error.

src/action/stat.rs
RainMustFall: Really, this approach seems better, [fixed](https://github.com/google/rrg/pull/34/commits/59a5e28575045d37b15281d808bcccc6569d1eaa).

src/action/stat.rs
RainMustFall: [Fixed](https://github.com/google/rrg/pull/34/commits/9277f329f74fe9fc09b42973594738928817d14e)

src/action/stat.rs
```diff
@@ -25,19 +24,20 @@ impl From<std::io::Error> for Error {
...
+    uid: u32,
+    gid: u32,
+    size: u64,
     access_time: Option<SystemTime>,
     modification_time: Option<SystemTime>,
     status_change_time: Option<SystemTime>,
-    blocks_number: Option<u32>,
-    block_size: Option<u32>,
-    represented_device: Option<u32>,
+    blocks_number: u32,
```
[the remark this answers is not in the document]
RainMustFall: [Fixed](https://github.com/google/rrg/pull/34/commits/9363640b1ca43815294fd5b5bb4d915b4c42aeeb)

src/action/stat.rs
```diff
@@ -25,19 +24,20 @@ impl From<std::io::Error> for Error {
...
 #[derive(Debug)]
 pub struct Response {
-    mode: Option<u64>,
-    inode: Option<u32>,
-    device: Option<u32>,
-    hard_links: Option<u32>,
-    uid: Option<u32>,
-    gid: Option<u32>,
-    size: Option<u64>,
+
```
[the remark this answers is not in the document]
RainMustFall: Fixed

src/action/stat.rs
```diff
@@ -0,0 +1,489 @@
...
+        path: original_path.to_owned(),
+
+        extended_attributes: vec![],
+    })
+}
+
+#[cfg(not(target_os = "linux"))]
+fn form_response(original_path: &PathBuf, destination: &PathBuf)
+                 -> Result<Response, std::io::Error> {
+    Ok(Response::default())
```
panhania, line 186: Looks like this lines prevents the code from compiling on Windows and macOS. Since you don't support these platforms, why not just return `Err` here?

RainMustFall: [Fixed](https://github.com/google/rrg/pull/34/commits/f19a8b879643b4dde2948d53a713ca5aa48cd12d)

PR #36, opened 2020-05-31
Add the `GetInstallDate` action

### LINKED ISSUE #14: Implement the `GetInstallDate` action
RRG should support the [`GetInstallDate`][install-action] action. It does not have any input arguments and should respond with an `uint64` timestamp.

[install-action]: https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/client/grr_response_client/client_actions/linux/linux.py#L185-L191

alex65536:
This pull request implements `GetInstallDate` action. Fixes #14.


src/action/install_date.rs
```diff
@@ -0,0 +1,267 @@
...
+use std::time::{SystemTime, Duration};
+
+use log::error;
+use crate::session::{self, Session};
+
+#[cfg(target_family = "unix")]
+use std::{fs, path::Path};
+
+#[cfg(target_os = "windows")]
+use winreg::RegKey;
```
M0rC1PanN, line 20: It can be in the function which uses it, I think.

alex65536: Good point :) Thanks, it's fixed now.

src/action/mod.rs
```diff
@@ -24,6 +24,7 @@ pub mod interfaces;
 pub mod metadata;
 pub mod startup;
 pub mod network;
+pub mod install_date;
```
panhania, line 27: Any ideas how to make this one word? :D I was trying to come up with something but failed :(

src/action/install_date.rs
```diff
@@ -0,0 +1,284 @@
...
+/// To obtain install date, it first checks the creation time of the root
+/// filesystem. Then, it searches for the installation logs and gets their
+/// modification date. It also checks for other files that were probably
+/// unchanged from the installation.
+///
+/// This function returns `None` in case of errors.
+#[cfg(target_os = "linux")]
+fn get_install_time() -> Option<SystemTime> {
+    // First, check the creation time of the root. This method works on
+    // Linux >= 4.11.
```
panhania, line 180: Could you make this comment more detailed (e.g. why doesn't it work on Linux pre-4.11)?

src/action/install_date.rs
```diff
@@ -0,0 +1,284 @@
...
+///
+/// To obtain install date, this function searches the corresponding value in
+/// Windows registry.
+///
+/// This function returns `None` in case of errors.
+#[cfg(target_os = "windows")]
+fn get_install_time() -> Option<SystemTime> {
+    use winreg::RegKey;
+
+    // Don't use winreg::enums::KEY_WOW64_64KEY since it breaks on Windows 2000
```
panhania, line 246: Could you clarify what does it mean to "break" in this context? Also: missing dot at the end.

src/action/install_date.rs
```diff
@@ -0,0 +1,284 @@
...
+    fn test_install_date() {
+        let mut session = session::test::Fake::new();
+        assert!(handle(&mut session, ()).is_ok());
+        assert_eq!(session.reply_count(), 1);
+        let response: &Response = session.reply(0);
+        let time = response.time.unwrap();
+
+        // We assume here that the tests won't be run for very old systems
+        // installed before 01.01.2000.
+        let lower_limit: Timestamp = "2000-01-01T00:00:00Z".parse().unwrap();
```
panhania, line 276: I am not sure if setting some arbitrary lower limit other than epoch makes sense (because it's hard to decide what should it be exactly).

src/action/install_date.rs
```diff
@@ -0,0 +1,284 @@
...
+    if let Ok(time) = time {
+        return Some(time);
+    }
+
+    // Then, try to detect filesystem creation time using dumpe2fs. This
+    // works only on ext2/3/4 filesystems.
+    let time = e2fs_utils::creation_time_from_dumpe2fs();
+    if let Some(time) = time {
+        return Some(time);
+    }
```
panhania, line 191: Same here, I think:
```rust
if time.is_some() {
    return time;
}
```
has a slightly less cognitive load.

src/action/install_date.rs
```diff
@@ -0,0 +1,284 @@
...
+    const CANDIDATES: [&str; 5] = [
+        "/var/log/installer/syslog",
+        "/var/log/installer/status",
+        "/root/install.log",
+        "/etc/hostname",
+        "/lost+found",
+    ];
+    if let Some(time) = get_modified_time(CANDIDATES.iter()) {
+        return Some(time);
+    }
```
panhania, line 212: Ditto.

src/action/install_date.rs
```diff
@@ -0,0 +1,284 @@
...
+    It: Iterator<Item = P>,
+{
+    for path in iter {
+        let time = fs::metadata(path).and_then(|metadata| metadata.modified());
+        if let Ok(time) = time {
+            return Some(time);
+        }
+    }
+
+    None
```
panhania, line 74: I think this could be simplified to this:

```rust
iter.filter_map(|path| fs::metadata(path).modified().ok()).next()
```

Although I am not sure if it is more readable ¯\\\_(ツ)\_/¯

src/action/install_date.rs
```diff
@@ -0,0 +1,284 @@
...
+    // Then, search for various files that were potentially modified only
+    // on installation:
+    // * /var/log/installer/: The location of installation logs for Debian-based
+    //   distributions.
+    // * /root/install.log: The location of installation logs for RHEL-based
+    //   distributions.
+    // * /etc/hostname: In most cases, we may assume that the hostname didn't
+    //   change after installation.
+    // * /lost+found: This method was used by Python version on GRR client, so
+    //   leaving it here.
```
panhania, line 202: I would backtick the paths here.

src/action/install_date.rs
alex65536: The dates that are much higher than epoch (like 2000-01-01) can potentially catch issues with wrong measurement units in the code, so I would prefer leaving it as is.

src/action/install_date.rs
alex65536: I don't know, this comment was just copied from GRR together with the idea of the implementation :)

src/action/mod.rs
alex65536: I also have no ideas whether it's possible to use a one-word name. Does it really matter? :)

src/action/install_date.rs
alex65536: Also, I'm not sure whether Rust support Windows 2000 or anyone is still using it. If not, this comment can be safely removed.

src/action/mod.rs
panhania: How about `insttime`? Unfortunately, I cannot say why it is better. I am just completely unable to justify this (which is generally not a good thing and making others follow this makes it even more terrible), but having single-word names for action modules just seems *right* and non-single-word names does not seem right.

I did some GitHub search and there is a non-zero amount of `insttime`s in peoples code, so its not completely unheard of. And it would nicely correspond to things like `uptime` or `boottime`.

src/action/install_date.rs
panhania: These issues should be covered by the upper-limit check, no? Of course not all of them, but then you could argue that some issues are not caught be the 2000.01.01 either. And the choice of it seems rather arbitrary.

Adding such assumptions, while can be seen as quite innocent here, sets a precedent for more serious cases.

PR #37, opened 2020-08-21
Create a helper trait for parsing protobuf enums.

Spawek:


src/session/parse_enum.rs
```diff
@@ -0,0 +1,63 @@
+use crate::session::UnknownEnumValueError;
+
+pub trait ProtoEnum<Proto> {
+    // Returns a default value for given Protobuf definition.
+    fn default() -> Proto;
+
+    // Returns value of the enum or None if the input i32 does not describe any know enum value.
+    // ::prost::Enumeration cannot be used instead, because it's not a trait.
```
panhania, line 8: Identifiers are not formatted properly.

src/session/parse_enum.rs
```diff
@@ -0,0 +1,63 @@
...
+    fn from_i32(val: i32) -> Option<Proto>;
+}
+
+// Maps the raw integer value to enum value or returns ParseError when the value cannot be mapped.
```
panhania, line 12: Ditto.

src/session/parse_enum.rs
```diff
@@ -0,0 +1,63 @@
...
+pub fn parse_enum<T: ProtoEnum<T>>(raw_enum_value: Option<i32>) -> Result<T, UnknownEnumValueError> {
```
panhania, line 13: Line too long. You can use `where` clasue to make it fit.

src/session/parse_enum.rs
```diff
@@ -0,0 +1,63 @@
...
+        fn default() -> TestEnum { TestEnum::Two }
+        fn from_i32(val: i32) -> Option<TestEnum> {
+            TestEnum::from_i32(val)
+        }
+    }
+
+    #[test]
+    fn parse_empty_enum_test() {
+        let empty: Option<i32> = None;
+        let parsed : TestEnum = parse_enum(empty).unwrap();
```
panhania, line 46: There should be no space before colon. Also, I don't think you need any type annotations here (and in these tests in general).

src/session/parse_enum.rs
```diff
@@ -0,0 +1,63 @@
...
+        assert_eq!(parsed, TestEnum::Two);
+    }
+
+    #[test]
+    fn parse_correct_enum_value_test() {
+        let parsed : TestEnum = parse_enum(Some(1)).unwrap();
+        assert_eq!(parsed, TestEnum::One);
```
panhania, line 53: Why not just `assert_eq!(parse_enum(Some(1)), Some(TestEnum::One))`?

src/session/parse_enum.rs
```diff
@@ -0,0 +1,63 @@
...
+        fn from_i32(val: i32) -> Option<TestEnum> {
+            TestEnum::from_i32(val)
+        }
+    }
+
+    #[test]
+    fn parse_empty_enum_test() {
+        let empty: Option<i32> = None;
+        let parsed : TestEnum = parse_enum(empty).unwrap();
+        assert_eq!(parsed, TestEnum::Two);
```
panhania, line 47: Why not just `assert_eq!(parse_enum(None), Some(TestEnum::Two))`

src/session/parse_enum.rs
```diff
@@ -0,0 +1,63 @@
...
+    use super::*;
+
+    #[derive(Debug, PartialEq, ::prost::Enumeration)]
+    pub enum TestEnum {
+        One = 1,
+        Two = 2,
+    }
+
+    impl ProtoEnum<TestEnum> for TestEnum {
+        fn default() -> TestEnum { TestEnum::Two }
```
panhania, line 37: I think it is not very intuitive that a value named `Two` is a default and makes the assertion below rather unobvious. I would use three variants: `Default = 0`, `One = 1` and `Two = 2`.

src/session/error.rs
```diff
@@ -179,3 +185,42 @@ impl From<MissingFieldError> for ParseError {
...
+pub struct UnknownEnumValueError {
+    pub enum_name: &'static str,
+    pub value: i32
+}
+
+impl UnknownEnumValueError {
+
+    /// Creates a new error indicating that a proto enum has a value for which the definition
+    /// is not known.
+    pub fn new(enum_name: &'static str, value: i32) -> UnknownEnumValueError {
```
panhania, line 200: In general, I don't think this constructor is useful. I think initializing this error always with the record syntax has better readability.

src/session/parse_enum.rs
```diff
@@ -0,0 +1,63 @@
...
+
+// Maps the raw integer value to enum value or returns ParseError when the value cannot be mapped.
+pub fn parse_enum<T: ProtoEnum<T>>(raw_enum_value: Option<i32>) -> Result<T, UnknownEnumValueError> {
+    match raw_enum_value {
+        Some(int_value) => match T::from_i32(int_value) {
+            Some(parsed_value) => Ok(parsed_value),
+            None => Err(UnknownEnumValueError::new(
+                std::any::type_name::<T>(),
+                int_value,
+            )),
```
panhania, line 20: This further proves my comment about using record syntax for initialization.

src/session/parse_enum.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    // Returns value of the enum or None if the input i32 does not describe any know enum value.
+    // ::prost::Enumeration cannot be used instead, because it's not a trait.
+    fn from_i32(val: i32) -> Option<Proto>;
+}
+
+// Maps the raw integer value to enum value or returns ParseError when the value cannot be mapped.
+pub fn parse_enum<T: ProtoEnum<T>>(raw_enum_value: Option<i32>) -> Result<T, UnknownEnumValueError> {
+    match raw_enum_value {
+        Some(int_value) => match T::from_i32(int_value) {
```
panhania, line 15: You could use the `ok_or_else` combinator.

panhania (changes requested):

src/session/error.rs
```diff
@@ -179,3 +185,42 @@ impl From<MissingFieldError> for ParseError {
         ParseError::malformed(error)
     }
 }
+
+/// An error type for situations where proto enum has a value for which the definition is not known.
```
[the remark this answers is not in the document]
Spawek: Fixed.

src/session/error.rs
```diff
@@ -179,3 +185,42 @@ impl From<MissingFieldError> for ParseError {
...
+#[derive(Debug)]
+pub struct UnknownEnumValueError {
+    pub enum_name: &'static str,
```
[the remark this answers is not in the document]
Spawek: Changed to "name".

src/session/error.rs
```diff
@@ -179,3 +185,42 @@ impl From<MissingFieldError> for ParseError {
...
+    pub value: i32
```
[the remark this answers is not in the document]
Spawek: Added docstrings.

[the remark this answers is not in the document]
Spawek: Added .

src/session/error.rs
```diff
@@ -179,3 +185,42 @@ impl From<MissingFieldError> for ParseError {
...
+}
+
+impl UnknownEnumValueError {
+
+    /// Creates a new error indicating that a proto enum has a value for which the definition
+    /// is not known.
+    pub fn new(enum_name: &'static str, value: i32) -> UnknownEnumValueError {
+        UnknownEnumValueError {
+            enum_name,
+            value
```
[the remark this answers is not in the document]
Spawek: This code was removed.

src/session/error.rs
Spawek: Removed.

src/session/error.rs
```diff
@@ -179,3 +185,42 @@ impl From<MissingFieldError> for ParseError {
...
+    }
+}
+
+impl std::error::Error for UnknownEnumValueError {
+
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        None
+    }
+}
+impl From<UnknownEnumValueError> for ParseError {
```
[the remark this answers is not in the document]
Spawek: Added.

src/session/error.rs
```diff
@@ -179,3 +185,42 @@ impl From<MissingFieldError> for ParseError {
...
+
+    fn from(error: UnknownEnumValueError) -> ParseError {
+        ParseError::UnknownEnumValue(error)
+    }
+}
```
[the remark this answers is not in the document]
Spawek: Added.

src/session/parse_enum.rs
Spawek: Fixed.

src/session/parse_enum.rs
Spawek: Fixed.

src/session/parse_enum.rs
Spawek: Changed to:
```
        Default = 1,
        Two = 2,
        Three = 3,
```

I don't want to use 0 as any value in here (and definitely not as default value). I don't want to have this test passing when 0 is returned by accident (which wouldn't be strange when playing with enums).

src/session/error.rs
```diff
@@ -89,6 +89,8 @@ impl From<ParseError> for Error {
 pub enum ParseError {
     /// An error occurred because the decoded proto message was malformed.
     Malformed(Box<dyn std::error::Error + Send + Sync>),
+    /// A protobuf had a value which is now known.
+    UnknownEnumValue(UnknownEnumValueError),
```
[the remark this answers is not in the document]
Spawek: Removed.

src/session/parse_enum.rs
```diff
@@ -0,0 +1,61 @@
+use crate::session::UnknownEnumValueError;
+
+pub trait ProtoEnum<Proto> {
+    /// Returns a default value for given Protobuf definition.
+    fn default() -> Proto;
```
panhania, line 5: This is covered by the standard `Default` trait.

src/session/parse_enum.rs
```diff
@@ -0,0 +1,61 @@
+use crate::session::UnknownEnumValueError;
+
+pub trait ProtoEnum<Proto> {
```
panhania, line 3: I am not sold on this trait. What is the problem is it trying to solve?

Spawek: You can see more context in the PR: https://github.com/google/rrg/pull/42. For multiple protobuf enums we need to:
- know the default value (which is not a property of enum itself, but of the structure which uses the enum),
- be able to parse it easily,
- be able to detect that the received value of the enum is not recognised and return an approrpiate error.

With example implementation:
```
impl ProtoEnum<ActionType> for ActionType {
    fn default() -> Self {
        FileFinderAction::default().action_type()
    }
    fn from_i32(val: i32) -> Option<Self> {
        ActionType::from_i32(val)
    }
}
```

You can parse the enum from proto just by:

```
parse_enum::<ActionType>(proto.action_type)?;
```

src/session/parse_enum.rs
Spawek: I think my previous answer covers this comment also.

src/session/parse_enum.rs
Spawek: Small thing to add: in most of cases you parse by just `parse_enum(proto.action_type)` - no type annotations are needed.

src/session/parse_enum.rs
Spawek: Fixed.

src/session/parse_enum.rs
Spawek: Fixed - however without `where` clause.

src/session/error.rs
```diff
@@ -180,25 +180,35 @@ impl From<MissingFieldError> for ParseError {
     }
 }
 
-/// An error type for situations where a given proto value is not supported.
+/// An error type for situations where proto enum has a value for which
+/// the definition is not known.
 #[derive(Debug)]
-pub struct UnsupportedValueError<T> {
-    /// A name of the field the value belongs to.
+pub struct UnknownEnumValueError {
```
panhania, line 186: I think `UnsupportedValueError` and `UnknownEnumValueError` are different, I don't think one should replace another.

src/session/parse_enum.rs
panhania: I don't think it does (if it does, then please, explain).

Spawek: The default value of any proto enum must be 0 ([link](https://developers.google.com/protocol-buffers/docs/proto3#:~:text=For%20enums%2C%20the%20default%20value,exact%20value%20is%20language%2Ddependent.)) In our case we care about the value to which we default when given proto enum is not set.

src/session/error.rs
Spawek: Brought back UnsupportedValueError - it was removed by accident during a merge.

panhania (approved):

PR #40, opened 2020-08-21
Create type for regex parsing errors.

Spawek:


src/session/error.rs
```diff
@@ -91,6 +91,8 @@ pub enum ParseError {
     Malformed(Box<dyn std::error::Error + Send + Sync>),
     /// An error occurred when decoding bytes of a proto message.
     Decode(prost::DecodeError),
+    /// An error occurred when parsing Vec<u8> to Regex.
+    RegexParse(RegexParseError)
```
panhania, line 95: I think this error is too specific to make it a separate value for the enum. `Malformed` should cover this case.

src/session/error.rs
```diff
@@ -179,3 +185,36 @@ impl From<MissingFieldError> for ParseError {
...
     }
 }
+
+#[derive(Debug)]
+pub struct RegexParseError {
+    /// Raw data of the string which could not be converted to Regex.
+    pub raw_data: Vec<u8>,
+
+    /// Error message caught during the conversion.
+    pub error_message: String,
```
panhania, line 195: Why store error message rather than the error itself?

panhania (approved):

PR #42, opened 2020-08-21
Create wrappers for the finder action protos.

Spawek:


src/action/finder/request.rs
```diff
@@ -0,0 +1,1309 @@
...
+type LiteralMatchMode =
+    rrg_proto::file_finder_contents_literal_match_condition::Mode;
+type ActionType = rrg_proto::file_finder_action::Action;
+type ConditionType = rrg_proto::file_finder_condition::Type;
+type XDevMode = rrg_proto::file_finder_args::XDev;
+type PathType = rrg_proto::path_spec::PathType;
+
+#[derive(Debug)]
+pub struct Request {
+    pub paths: Vec<String>,
```
panhania, line 39: There is a `Path` type that should be used to represent paths.

src/action/finder/request.rs
```diff
@@ -0,0 +1,1309 @@
...
+#[derive(Debug)]
+pub enum Action {
+    Stat(StatActionOptions),
+    Hash(HashActionOptions),
+    Download(DownloadActionOptions),
+}
+
+#[derive(Debug)]
+pub struct StatActionOptions {
+    pub resolve_links: bool,
```
panhania, line 57: How does it interact with `follow_links` of `Request`?

src/action/finder/request.rs
```diff
@@ -0,0 +1,1309 @@
...
+type ActionType = rrg_proto::file_finder_action::Action;
+type ConditionType = rrg_proto::file_finder_condition::Type;
+type XDevMode = rrg_proto::file_finder_args::XDev;
+type PathType = rrg_proto::path_spec::PathType;
+
+#[derive(Debug)]
+pub struct Request {
+    pub paths: Vec<String>,
+    pub path_type: PathType,
+    pub action: Option<Action>,
```
panhania, line 41: Why is it optional?

src/action/finder/request.rs
```diff
@@ -0,0 +1,1309 @@
...
+    pub xor_out_key: u32,
+}
+
+impl TryFrom<rrg_proto::FileFinderAction> for Action {
+    type Error = ParseError;
+
+    fn try_from(
+        proto: rrg_proto::FileFinderAction,
+    ) -> Result<Self, Self::Error> {
+        // FileFinderAction::action_type defines which action will be performed.
```
panhania, line 129: Missing markdown backticks to quote an identifier.

src/action/finder/request.rs
Spawek: `StatActionOptions::resolve_links` is a property of "Stat" and is used when doing "Stat" on found files.
`Request::follow_links` is used when doing recursive scans.

src/action/finder/request.rs
Spawek: Added!

src/action/finder/request.rs
Spawek: Removed optional

src/action/finder/request.rs
Spawek: It's not really a path in this place - it can contain alternatives (e.g. "/home/{a,b,c/d,..}/test"), so it's simpler to use string to process it in early stages. I'm changing it to PathBuf after the preprocessing part.

src/action/finder/request.rs
```diff
@@ -0,0 +1,1405 @@
...
+            action: Some(FileFinderAction {
+                action_type: Some(ActionType::Stat as i32),
+                ..Default::default()
+            }),
+            ..Default::default()
+        })
+        .unwrap();
+
+        match request.action {
+            Action::Stat(options) => {
```
panhania, line 616: Prefer to use `if let` to reduce nesting (here and below).

src/action/finder/request.rs
```diff
@@ -0,0 +1,1405 @@
...
+            let length = options.length();
+            let xor_in_key = options.xor_in_key();
+            let xor_out_key = options.xor_out_key();
+            let mode =
+                MatchMode::from(parse_enum::<LiteralMatchMode>(options.mode)?);
+
+            if options.literal.is_none() {
+                return Ok(vec![]);
+            }
+            let literal = options.literal.unwrap();
```
panhania, line 476: Ditto.

src/action/finder/request.rs
```diff
@@ -0,0 +1,1405 @@
...
+}
+
+fn get_modification_time_conditions(
+    proto: Option<FileFinderModificationTimeCondition>,
+) -> Result<Vec<Condition>, ParseError> {
+    Ok(match proto {
+        Some(options) => {
+            let mut conditions: Vec<Condition> = vec![];
+            match options.min_last_modified_time {
+                Some(micros) => {
```
panhania, line 283: `if let` (here and below).

src/action/finder/request.rs
```diff
@@ -0,0 +1,1405 @@
...
+            LiteralMatchMode::AllHits => MatchMode::AllHits,
+        }
+    }
+}
+
+fn get_modification_time_conditions(
+    proto: Option<FileFinderModificationTimeCondition>,
+) -> Result<Vec<Condition>, ParseError> {
+    Ok(match proto {
+        Some(options) => {
```
panhania, line 280: Nesting in these functions is a bit too ridiculous, try to make it less deep (e.g. by using early returns).

src/action/finder/request.rs
```diff
@@ -0,0 +1,1405 @@
...
+    MinAccessTime(std::time::SystemTime),
+    MaxAccessTime(std::time::SystemTime),
+    MinInodeChangeTime(std::time::SystemTime),
+    MaxInodeChangeTime(std::time::SystemTime),
+    MinSize(u64),
+    MaxSize(u64),
+    ExtFlagsLinuxBitsSet(u32),
+    ExtFlagsLinuxBitsUnset(u32),
+    ExtFlagsOsxBitsSet(u32),
+    ExtFlagsOsxBitsUnset(u32),
```
panhania, line 90: Original file-finder groups these into one condition. Why did you decide to split these? Usually enums should be fairly small, otherwise they encourage doing wildcard pattern matching.

src/action/finder/request.rs
panhania: Then you have an incorrectly named field.

src/action/finder/request.rs
```diff
@@ -0,0 +1,1405 @@
...
+    rrg_proto::file_finder_contents_literal_match_condition::Mode;
+type ActionType = rrg_proto::file_finder_action::Action;
+type ConditionType = rrg_proto::file_finder_condition::Type;
+type XDevMode = rrg_proto::file_finder_args::XDev;
+type PathType = rrg_proto::path_spec::PathType;
+
+#[derive(Debug)]
+pub struct Request {
+    pub paths: Vec<String>,
+    pub path_type: PathType,
```
panhania, line 40: Is this used?

panhania (changes requested):

src/action/finder/request.rs
```diff
@@ -0,0 +1,1405 @@
...
+            let bytes_after = options.bytes_after();
+            let start_offset = options.start_offset();
+            let length = options.length();
+            let mode =
+                MatchMode::from(parse_enum::<RegexMatchMode>(options.mode)?);
+
+            if options.regex.is_none() {
+                return Ok(vec![]);
+            }
+            let regex = parse_regex(options.regex.unwrap())?;
```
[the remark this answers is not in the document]
Spawek: Fixed!

src/action/finder/request.rs
Spawek: Fixed!

src/action/finder/request.rs
Spawek: Done!

src/action/finder/request.rs
Spawek: Done!

src/action/finder/request.rs
Spawek: Reduced it by using 2 "if let" expressions

src/action/finder/request.rs
Spawek: I was considering it, but I believe that storing 4 Optional<Condition> is worse than adding 4 values to this enum. I want to keep those conditions as simple as possible.

src/action/finder/request.rs
Spawek: This field is a copy of FileFinderArgs::paths (which has the same type). I think that renaming it wouldn't be helpful, because idea of this field is being a `path`.

src/action/finder/request.rs
Spawek: It is used e.g. for setting the type to "Registry".

src/action/finder/request.rs
```diff
@@ -0,0 +1,1355 @@
...
+            error: e,
+        }
+        .into()),
+    }
+}
+
+fn get_contents_regex_match_condition(
+    proto: Option<FileFinderContentsRegexMatchCondition>,
+) -> Result<Vec<Condition>, ParseError> {
+    if let Some(options) = proto {
```
panhania, line 382: You can reduce nesting by inverting the condition:

```rust
let options = match proto {
    Some(options) => options,
    None => return Ok(vec!())`,
}
```

src/action/finder/request.rs
```diff
@@ -0,0 +1,1355 @@
...
+        };
+        return Ok(vec![Condition::ContentsRegexMatch(ret)]);
+    }
+    Ok(vec![])
+}
+
+fn get_contents_literal_match_condition(
+    proto: Option<FileFinderContentsLiteralMatchCondition>,
+) -> Result<Vec<Condition>, ParseError> {
+    if let Some(options) = proto {
```
panhania, line 412: Ditto.

panhania (changes requested):

src/action/finder/request.rs
Spawek: Removed `path_type` from code.

src/action/finder/request.rs
```diff
@@ -0,0 +1,1355 @@
...
+        let bytes_before = options.bytes_before();
+        let bytes_after = options.bytes_after();
+        let start_offset = options.start_offset();
+        let length = options.length();
+        let mode = MatchMode::from(parse_enum::<RegexMatchMode>(options.mode)?);
+
+        let regex = match options.regex {
+            None => {
+                return Ok(vec![]);
+            }
```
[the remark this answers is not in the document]
Spawek: Fixed.

src/action/finder/request.rs
Spawek: Done!

src/action/finder/request.rs
Spawek: Fixed!

panhania (approved):

PR #43, opened 2020-08-21
Add time conversion utilities.

Spawek:


src/session/error.rs
```diff
@@ -91,6 +91,9 @@ pub enum ParseError {
     Malformed(Box<dyn std::error::Error + Send + Sync>),
     /// An error occurred when decoding bytes of a proto message.
     Decode(prost::DecodeError),
+    /// An error occurred when converting time micros from proto
+    /// to `std::time::SystemTime`.
+    TimeMicrosConversion(TimeMicrosConversionError)
```
panhania, line 96: Likewise, I don't think we should add too much variants into the `ParseError`. `TmeMicrosConversion` falls under `Malformed` IMO.

src/session/time.rs
```diff
@@ -0,0 +1,16 @@
...
+    micros: u64,
+) -> Result<std::time::SystemTime, TimeMicrosConversionError> {
+    let result = std::time::UNIX_EPOCH
+        .checked_add(std::time::Duration::from_micros(micros));
+
+    if result.is_none() {
+        return Err(TimeMicrosConversionError { micros });
+    }
+
+    Ok(result.unwrap())
```
panhania, line 15: Avoid using `Option::unwrap` in any code. Here, this should be easily expressible with `Option::ok_or`.

panhania (changes requested):

src/session/error.rs
Spawek: Removed the variant - merged it into `Malformed`.

src/session/time.rs
Spawek: `ok_or` makes that code much better! Thanks for the suggestion!

src/session/error.rs
```diff
@@ -90,7 +90,7 @@ pub enum ParseError {
     /// An error occurred because the decoded proto message was malformed.
     Malformed(Box<dyn std::error::Error + Send + Sync>),
     /// An error occurred when decoding bytes of a proto message.
-    Decode(prost::DecodeError),
```
panhania, line 93: Why delete this trailing comma...?

src/session/error.rs
```diff
@@ -202,3 +202,32 @@ impl<T: Debug> std::error::Error for UnsupportedValueError<T> {
         None
     }
 }
+
+/// An error type for situations where time micros cannot be converted
+/// to `std::time::SystemTime`.
+#[derive(Debug)]
+pub struct TimeMicrosConversionError {
+    /// Time micros value causing the conversion error.
+    pub micros: u64
```
panhania, line 211: Oh, and here is a missing trailing comma.

src/session/error.rs
Spawek: Added it back.

panhania (approved):

PR #44, opened 2020-09-07
Add a function translating globs to regexes.

Spawek:


src/action/client_side_file_finder/glob_to_regex.rs
```diff
@@ -0,0 +1,86 @@
...
+// in the LICENSE file or at https://opensource.org/licenses/MIT.
+
+use crate::session::RegexParseError;
+use regex::Regex;
+
+/// Converts glob expression into a Rust Regex.
+/// E.g. "asd*[!123]??" will be converted into "asd.*[^123]..".
+/// This implementation is a Rust port of the cpython code:
+/// https://github.com/python/cpython/blob/2.7/Lib/fnmatch.py
+fn glob_to_regex(pat: &str) -> Result<Regex, RegexParseError> {
```
panhania, line 13: I am not sure if I like this. In Python we use `fnmatch` to translate glob to a regex and then use this for matching because it avoids implementing globbing mechanism ourselves. However, if we have to implement `fnmatch` ourselves, it defeats the whole purpose of its existence and we can implement globbing directly, without unnecessarily going through a regex.

Spawek: In RRG we have to implement globbing ourselves. Part of globbing implementation is deciding if a directory/file name matches a pattern. I see 2 ways of doing that:
1) Convert the glob to regex::Regex and use the regex library to check if name matches the pattern.
2) Implement a functionality `matches_glob(text, glob) -> bool`.

I believe that difficulty of both solutions is similar (because of edge-cases like unmatched parenthesis, characters escaping etc). In this PR I decided to implement conversion to regex as we had a working implementation from cpython that was already used in GRR and I don't expect it to cause any problems.

src/action/finder/glob_to_regex.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_glob_to_regex() {
+        // Original cpython tests:
```
panhania, line 71: Rather than making comments, split the test case into multiple cases.

src/action/finder/glob_to_regex.rs
```diff
@@ -0,0 +1,86 @@
...
+        assert_eq!(glob_to_regex("[abc]").unwrap().as_str(), "[abc]");
+        assert_eq!(glob_to_regex("[]]").unwrap().as_str(), "[]]");
+        assert_eq!(glob_to_regex("[!x]").unwrap().as_str(), "[^x]");
+        assert_eq!(glob_to_regex("[^x]").unwrap().as_str(), r"[\^x]");
+        assert_eq!(glob_to_regex("[x").unwrap().as_str(), r"\[x");
+
+        // Additional tests:
+        assert_eq!(glob_to_regex("[a]]").unwrap().as_str(), r"[a]\]");
+        assert_eq!(glob_to_regex(r"[\\]\\").unwrap().as_str(), r"[\\]\\");
+        assert_eq!(glob_to_regex("ąźć").unwrap().as_str(), "ąźć");
```
panhania, line 84: Same here, I would split this into three separate case with descriptive names that clearly formulate what you are testing for, e.g.

```rust
fn test_glob_to_regex_unice() {
    // ...
}
```

src/action/finder/glob_to_regex.rs
```diff
@@ -0,0 +1,86 @@
+// Copyright 2020 Google LLC
+//
+// Use of this source code is governed by an MIT-style license that can be found
+// in the LICENSE file or at https://opensource.org/licenses/MIT.
+
+use crate::session::RegexParseError;
+use regex::Regex;
+
+/// Converts glob expression into a Rust Regex.
+/// E.g. "asd*[!123]??" will be converted into "asd.*[^123]..".
```
panhania, line 10: Prefer to use examples that are verified when you run tests, e.g. like [here][1] or [here][2].

[1]: https://github.com/google/rrg/blob/40447d159b1d65ff0a7a5056503509a780a4aba1/src/metadata.rs#L58-L85
[2]: https://github.com/google/rrg/blob/40447d159b1d65ff0a7a5056503509a780a4aba1/src/fs/mod.rs#L34-L77

src/action/finder/glob_to_regex.rs
```diff
@@ -0,0 +1,86 @@
...
+/// This implementation is a Rust port of the cpython code:
+/// https://github.com/python/cpython/blob/2.7/Lib/fnmatch.py
+fn glob_to_regex(pat: &str) -> Result<Regex, RegexParseError> {
```
panhania, line 13: GRR uses `str` to represent paths and this is a very big mistake that was made, in RRG we should make things *right* from the beginning. We should really think hard how to tackle this problem.

src/action/finder/glob_to_regex.rs
```diff
@@ -0,0 +1,86 @@
...
+                    stuff = String::from(r"\") + &stuff;
+                }
+                res = format!("{}[{}]", res, stuff);
+            }
+        } else {
+            res = res + &regex::escape(&c.to_string());
+        }
+    }
+
+    // cpython version produces output with escaped slashes, which is not desired here.
```
panhania, line 54: CPython*.

panhania (changes requested):

src/action/client_side_file_finder/glob_to_regex.rs
panhania: > Part of globbing implementation is deciding if a directory/file name matches a pattern. 

This is just the way GRR does it and it is not the most efficient of doing things. The proper way would be to split path into proper glob components and do traversal and matching ourselves. I understand that the solution presented here is a sort of acceptable temporary hack and maybe I am willing to accept it, but this is definitely not the way things should be implemented in a production-ready RRG agent.

panhania: Also, have you verified all the licensing implications? In GRR was just copy-and-pasted from the standard library without any changes. Here, I am not sure what are the legal requirements.

Spawek: This implementation was only inspired by the CPython code - I changed the comment to reflect it.

I don't agree that using Regex can't be a very efficient way of doing things. The plan is to do the traversal ourselves (you can see a working version here: https://github.com/Spawek/rrg/blob/dev/src/action/finder/action.rs - it's not ready for PR yet) and use Regex just to match directory names.

src/action/finder/glob_to_regex.rs
Spawek: Added an example. I agree it's much better that way.

src/action/finder/glob_to_regex.rs
Spawek: Changed

src/action/finder/glob_to_regex.rs
Spawek: I believe that those tests are self-explanatory. I removed those comments.

src/action/finder/glob_to_regex.rs
Spawek: I believe I understand the problem (at least partially). I designed the code to make a path transition to other data types as trivial as possible.

In the current version of the proto definition we use the input FileFinderArgs::paths contains `std::string::String` already` and I believe this is the first problem we should address.

I'm more than happy to meet to consider options on using different types for representing paths.

panhania (approved):

PR #45, opened 2020-09-11
Add function for expanding path groups.

Spawek:
For example, given path `foo/{bar,baz}/{quux,norf}` it will yield `foo/bar/quux`, `foo/bar/norf`, `foo/baz/quux`, `foo/baz/norf`.


src/action/mod.rs
```diff
@@ -28,6 +28,7 @@ pub mod timeline;
 pub mod network;
 pub mod insttime;
 pub mod memsize;
+pub mod client_side_file_finder;
```
panhania, line 31: Too long, I like module names to be one word. How about `finder`? It's not like you can have server-side one on the client. And then the "file" part can be omitted, because what else would you search for?

src/action/client_side_file_finder/mod.rs
```diff
@@ -0,0 +1,8 @@
+// Copyright 2020 Google LLC
+//
+// Use of this source code is governed by an MIT-style license that can be found
+// in the LICENSE file or at https://opensource.org/licenses/MIT.
+
+//! Handler for `client side file finder` action.
+
+mod expand_groups;
```
panhania, line 8: Same. Having a function-specific module seems weird. For now I would put it into something like `util` and then, once it grows enough, split it into more categories.

src/action/client_side_file_finder/expand_groups.rs
```diff
@@ -0,0 +1,71 @@
...
+        chunks.push(path[m.start()+1..m.end()-1].split(',').map(|x| x.to_owned()).collect());
+        offset = m.end();
+    }
+    chunks.push(vec![path[offset..].to_owned()]);
+
+    join_chunks(chunks)
+}
+
+fn join_chunks(chunks : Vec<Vec<String>>) -> Vec<String> {
+    if chunks.is_empty(){
```
panhania, line 40: Missing space before brace.

src/action/client_side_file_finder/expand_groups.rs
```diff
@@ -0,0 +1,71 @@
...
+    ret
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn expand_groups_test()
+    {
```
panhania, line 64: Curly brace should go to the function line.

src/action/client_side_file_finder/expand_groups.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    ret
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn expand_groups_test()
```
panhania, line 63: So far the convention is to use `test_*` rather than `*_test` (so that it reads like a proper sentence).

src/action/client_side_file_finder/expand_groups.rs
```diff
@@ -0,0 +1,71 @@
...
+    let mut ret = chunks[0].clone();
+    for chunk in chunks.into_iter().skip(1){
+        let mut new_ret = vec![];
+        for prefix in ret {
+            for suffix in &chunk {
+                new_ret.push(prefix.to_owned() + suffix);
+            }
+        }
+        ret = new_ret;
+    }
```
panhania, line 53: I don't understand (some kind of cartesian product?) and I should not try to—the docstring should be everything I need.

src/action/client_side_file_finder/expand_groups.rs
```diff
@@ -0,0 +1,71 @@
...
+///
+/// Differences from GRR:
+/// - support for empty value in the 1st and 2nd alternative e.g.:
+///   - `/{,asd}`, `/{asd,}` are resolved here, but not in GRR
+///   - `/{a,b,}` are resolved both in GRR and in here
+/// - support for a single alternative e.g.:
+///   - `/{asd}` is resolved here, but not in GRR
+pub fn expand_groups(path: &str) -> Vec<String> {
+    lazy_static! {
+        static ref RE : Regex = Regex::new(r"\{[^\{]*?\}").unwrap();
```
panhania, line 23: Naming regex variables `RE` is as useful as naming integer variables `integer`.

src/action/client_side_file_finder/expand_groups.rs
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    let mut offset = 0;
+    let mut chunks : Vec<Vec<String>> = vec![];
+
+    for m in RE.find_iter(path){
+        chunks.push(vec![path[offset..m.start()].to_owned()]);
+        chunks.push(path[m.start()+1..m.end()-1].split(',').map(|x| x.to_owned()).collect());
```
panhania, line 31: There is *a lot* of allocations flying around. Are you sure it cannot be avoided?

panhania (changes requested):

src/action/client_side_file_finder/mod.rs
Spawek: It already grew significantly on my `dev` branch, so I split it into categories. I believe there is a beauty in such a split as this module provides only `fn expand_groups(path: &str) -> Vec<String>` functionality and reader probably will never have to look at this module unless interested in the group expansion.

`Do one thing, and do it well`

src/action/client_side_file_finder/expand_groups.rs
Spawek: Fixed.

src/action/client_side_file_finder/expand_groups.rs
Spawek: Fixed all formating problems using `rustfmt`.

src/action/client_side_file_finder/expand_groups.rs
Spawek: Changed.

src/action/client_side_file_finder/expand_groups.rs
Spawek: Removed the allocations by changing the type of `chunks` to `Vec<Vec<&str>>`.

src/action/mod.rs
Spawek: Renamed to `finder` as discussed offline.

src/action/finder/groups.rs
```diff
@@ -0,0 +1,114 @@
...
+/// the order of appearance, where exactly 1 alternative is used from each
+/// chunk.
+/// The order of alternatives is defined as if there were generated by `for`
+/// loops nested in order of chunks.
+///
+/// Given the example input this function outputs:
+/// ["/home/spawek/.gitignore",
+///  "/home/spawek/.hgignore",
+///  "/home/hanuszczak/.gitignore",
+///  "/home/hanuszczak/.hgignore"].
```
panhania, line 55: This should use the doccomments example syntax.

src/action/finder/groups.rs
```diff
@@ -0,0 +1,114 @@
...
+fn join_chunks(chunks: Vec<Vec<&str>>) -> Vec<String> {
+    if chunks.is_empty() {
+        return vec![];
+    }
+
+    let mut ret: Vec<String> =
+        chunks[0].iter().map(|x| (*x).to_owned()).collect();
```
panhania, line 62: You can use:

```rust
let first = match chunks.first() {
    Some(first) => first,
    None => return vec!(),
}
```

to avoid writing direct indexing (that can be potentially dangerous).

src/action/finder/groups.rs
```diff
@@ -0,0 +1,114 @@
...
+        );
+        assert_eq!(
+            expand_groups("/some/{1,}a/{3,}a"),
+            vec!["/some/1a/3a", "/some/1a/a", "/some/a/3a", "/some/a/a"]
+        );
+        assert_eq!(expand_groups("/some/{123}"), vec!["/some/123"]);
+    }
+
+    #[test]
+    fn test_join_chunks() {
```
panhania, line 99: There are no tests for corner cases (e.g. what if one of the inputs is empty, what if there are no chunks at all etc.).

panhania (changes requested):

src/action/finder/groups.rs
Spawek: Added doccomment with `ignore` attribute as this function is private.

src/action/finder/groups.rs
Spawek: Done!

src/action/finder/groups.rs
Spawek: I don't think testing the case with one of the inputs being empty is not necessary as this is a private function and the regex used for fetching alternatives doesn't allow it.

Added that and other tests anyway.

PR #48, opened 2020-12-17
Add support for the hash action in CSFF.

Spawek:


src/action/finder/hash.rs
```diff
@@ -0,0 +1,190 @@
...
+use std::cmp::min;
+use std::fs::File;
+use std::io::Read;
+
+/// Hashes data writen to it using SHA-1, SHA-256 and MD5 algorithms.
+struct Hasher {
+    sha1: crypto::sha1::Sha1,
+    sha256: crypto::sha2::Sha256,
+    md5: crypto::md5::Md5,
+    num_bytes: usize,
```
panhania, line 16: I would prefer `total_byte_count`, `num_bytes` is very awkward.

src/action/finder/hash.rs
```diff
@@ -0,0 +1,190 @@
...
+        }
+    }
+}
+
+impl std::io::Write for Hasher {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        self.sha1.input(buf);
+        self.sha256.input(buf);
+        self.md5.input(buf);
+        self.num_bytes = self.num_bytes + buf.len();
```
panhania, line 35: `+=`

src/action/finder/hash.rs
```diff
@@ -0,0 +1,190 @@
...
+
+        Ok(buf.len())
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        Ok(())
+    }
+}
+
+pub fn hash(entry: &Entry, config: &HashActionOptions) -> Option<HashEntry> {
```
panhania, line 45: Public functions should be documented (private ones preferably as well).

src/action/finder/hash.rs
```diff
@@ -0,0 +1,190 @@
...
+    match config.oversized_file_policy {
+        OversizedFilePolicy::Skip => {
+            if entry.metadata.len() > config.max_size {
+                return None;
+            }
+        }
+        OversizedFilePolicy::HashTruncated => {}
```
panhania, line 52: Prefer `()` over `{}`. Reason: for me syntax of curly braces is always: "open on the same line, close on a newline":

```rust
foo {
    bar();
    baz();
}
```

Thus, if the body is empty, to stick to this rule it should look like:

```rust
foo {
}
```

But this is ugly, so I prefer to avoid using empty curly braces altogether.

src/action/finder/hash.rs
```diff
@@ -0,0 +1,190 @@
...
+        Ok(file) => file.take(config.max_size),
+        Err(err) => {
+            warn!(
+                "failed to open file: {}, error: {}",
+                entry.path.display(),
+                err
+            );
+            return None;
+        }
+    };
```
panhania, line 65: https://github.com/google/rrg/blob/08919801c64919f5b7277961d51c3ed2f2a8c9b4/macro/src/lib.rs#L81-L148

src/action/finder/hash.rs
```diff
@@ -0,0 +1,190 @@
...
+                entry.path.display(),
+                &err
+            );
+        }
+    }
+
+    Some(HashEntry {
+        sha1: Some(hasher.sha1.result_str().as_bytes().to_vec()),
+        sha256: Some(hasher.sha256.result_str().as_bytes().to_vec()),
+        md5: Some(hasher.md5.result_str().as_bytes().to_vec()),
```
panhania, line 93: Uhm, should it really use the hexadecimal representation?

src/action/finder/hash.rs
```diff
@@ -0,0 +1,190 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use rrg_proto::file_finder_hash_action_options::OversizedFilePolicy;
+
+    #[test]
+    fn test_hash_values() {
+        let test_string = "some_test_data";
```
panhania, line 110: You use this only once, why not inline this instead?

src/action/finder/hash.rs
```diff
@@ -0,0 +1,190 @@
...
+        assert_eq!(
+            result.sha256.unwrap(),
+            "d76d85adca8afad205edebc11f9b5086bca75acb512a748bc79660e1346af546"
+                .as_bytes()
+                .to_vec()
+        );
+        assert_eq!(
+            result.md5.unwrap(),
+            "e091b6f1a233049d22d2807fa8086f3f".as_bytes().to_vec()
+        );
```
panhania, line 143: You should be able to compare against raw byte literals.

panhania (changes requested):

src/action/finder/hash.rs
Spawek: Changed to `=> ()` as rustdocs are using that convention too, so it seems to be a standard.

src/action/finder/hash.rs
Spawek: I'm exiting the function on error, so `ack!` is not very helpful in this case.

src/action/finder/hash.rs
Spawek: Inlined!

src/action/finder/hash.rs
Spawek: Thanks for spotting this bug. I changed the code to return a vector of resulting bytes.

src/action/finder/hash.rs
Spawek: Makes sense. Changed!

src/action/finder/hash.rs
Spawek: Changed.

src/action/finder/hash.rs
Spawek: Added docs.

src/action/finder/hash.rs
Spawek: After an offline discussion used `ack!{...}`?;

src/action/finder/hash.rs
Spawek: Thanks for spotting this bug. I changed the code to return a vector of resulting bytes.

src/action/finder/hash.rs
Spawek: Thanks for spotting this bug. I changed the code to return a vector of resulting bytes.

src/action/finder/hash.rs
```diff
@@ -0,0 +1,201 @@
...
+        pecoff_md5: None,
+        pecoff_sha256: None,
+        signed_data: vec![],
+        num_bytes: Some(hasher.total_byte_count),
+        source_offset: None,
+    })
+}
+
+fn result_vec<T: Digest>(digest: &mut T) -> Vec<u8> {
+    let mut vec = repeat(0).take(digest.output_bytes()).collect::<Vec<u8>>();
```
panhania, line 99: This must be the most convoluted way of writing `vec![0; digest.output_bytes()]` I have seen.

src/action/finder/hash.rs
```diff
@@ -0,0 +1,201 @@
...
+        Ok(buf.len())
+    }
+
+    fn flush(&mut self) -> std::io::Result<()> {
+        Ok(())
+    }
+}
+
+/// Performs `hash` action on the file in `entry` and returns the result to be reported in case of success.
+pub fn hash(entry: &Entry, config: &HashActionOptions) -> Option<HashEntry> {
```
panhania, line 52: Is there a deeper reason why we need to work with raw protos (that has a lot of unused fields) instead of creating proper wrappers?

src/action/finder/hash.rs
```diff
@@ -0,0 +1,201 @@
...
+            }
+        }
+        OversizedFilePolicy::HashTruncated => {}
+    };
+
+    let file = ack! {
+        File::open(&entry.path),
+        error: "failed to open file: {}", entry.path.display()
+    }?;
+    let mut file = file.take(config.max_size);
```
panhania, line 66: You can just chain `take`, no need to introduce a variable only to shadow it on the next line.

src/action/finder/hash.rs
Spawek: Actually idea of doing it that way comes from the crypto library. `vec![0; digest.output_bytes()]` looks way better.

src/action/finder/hash.rs
Spawek: Added a wrapper (`FileHash`).

src/action/finder/hash.rs
Spawek: I believe `ack! {...}?.take()` is a bit too much for 1 statement. If you disagree I'm happy to change it.

panhania (approved):

PR #49, opened 2021-01-25
Download action support in Client Side File Finder.

Spawek:


src/action/finder/download.rs
```diff
@@ -0,0 +1,263 @@
...
+
+fn chunks<R: std::io::Read>(reader: R, chunk_size: u64) -> Chunks<R> {
+    Chunks {
+        bytes: reader.bytes(),
+        chunk_size,
+    }
+}
+
+#[derive(Debug)]
+pub struct Chunks<R> {
```
panhania, line 68: Does it need to be public?

src/action/finder/download.rs
```diff
@@ -0,0 +1,263 @@
...
+    bytes: std::io::Bytes<R>,
+    chunk_size: u64,
+}
```
panhania, line 71: Missing documentation.

src/action/finder/download.rs
```diff
@@ -0,0 +1,263 @@
...
+
+            if ret.len() == self.chunk_size as usize {
+                return Some(Ok(ret));
+            }
+        }
+        if !ret.is_empty() {
+            return Some(Ok(ret));
+        }
+
+        return None;
```
panhania, line 93: Why not use something like [`read_exact`]?

[`read_exact`]: https://doc.rust-lang.org/std/io/trait.Read.html#method.read_exact

panhania (changes requested):
Some of the comments from the review in #48 are applicable here as well.

src/action/finder/download.rs
Spawek: From docs:
`If this function encounters an "end of file" before completely filling the buffer, it returns an error of the kind ErrorKind::UnexpectedEof. The contents of buf are unspecified in this case.`

src/action/finder/download.rs
Spawek: Yes - `Chunks` is visible in the public interface (`Response`).

src/action/finder/download.rs
Spawek: Added docs.

Spawek: Changed the names of the fields in this structure.

src/action/finder/chunks.rs
```diff
@@ -0,0 +1,170 @@
...
+    overlap_data: Vec<u8>,
+}
+
+pub struct GetFileChunksConfig {
+    /// Number of bytes skipped from the beginning of the file.
+    pub start_offset: u64,
+    /// Maximum number of bytes read from the file.
+    pub max_read_bytes: u64,
+    /// Desired number of bytes in chunks. Only the last chunk can be smaller
+    /// than the `bytes_per_chunk`
```
panhania, line 23: Missing dot.

src/action/finder/chunks.rs
```diff
@@ -0,0 +1,170 @@
...
+    pub bytes_per_chunk: u64,
+    /// A number of bytes that the next chunk will share with the previous one.
+    pub overlap_bytes: u64,
+}
+
+pub fn get_file_chunks(
+    path: &Path,
+    config: &GetFileChunksConfig,
+) -> Option<Chunks<BufReader<Take<File>>>> {
```
panhania, line 32: Missing documentation for a public function.

src/action/finder/chunks.rs
```diff
@@ -0,0 +1,170 @@
...
+            overlap_bytes: config.overlap_bytes,
+        },
+    ))
+}
+
+fn open_file(
+    path: &Path,
+    offset: u64,
+    max_size: u64,
+) -> Option<Take<File>> {
```
panhania, line 48: This fits single line, no?

src/action/finder/chunks.rs
```diff
@@ -0,0 +1,170 @@
...
+        BufReader::new(file),
+        ChunksConfig {
+            bytes_per_chunk: config.bytes_per_chunk,
+            overlap_bytes: config.overlap_bytes,
+        },
+    ))
+}
+
+fn open_file(
+    path: &Path,
```
panhania, line 45: Prefer to use `AsRef<Path>` (as all standard library functions do).

src/action/finder/chunks.rs
```diff
@@ -0,0 +1,170 @@
...
+                );
+                return None;
+            }
+            Some(f.take(max_size))
+        }
+        Err(err) => {
+            warn!("failed to open file: {}, error: {}", path.display(), err);
+            None
+        }
+    }
```
panhania, line 65:
```rust
let file = match File::open(path) {
    Ok(file) => file,
    Err(err) => ...,
}

...
```
(to reduce nesting)

src/action/finder/chunks.rs
```diff
@@ -0,0 +1,170 @@
...
+        assert_eq!(chunks.next().unwrap().unwrap(), vec![1, 2, 3]);
+        assert_eq!(chunks.next().unwrap().unwrap(), vec![3, 4, 5]);
+        assert!(chunks.next().is_none());
+    }
+
+    #[test]
+    fn test_get_file_chunks_basic_use_case() {
+        let tempdir = tempfile::tempdir().unwrap();
+        let path = tempdir.path().join("f");
+        std::fs::write(&path, [1,2,3,4,5,6]).unwrap();
```
panhania, line 156: Spaces after comma. But I think I would use byte literals here (`b"foobar"`).

src/action/finder/chunks.rs
```diff
@@ -0,0 +1,170 @@
...
+
+        let mut chunks = get_file_chunks(&path, &GetFileChunksConfig{
```
panhania, line 158: Space before opening brace.

src/action/finder/download.rs
```diff
@@ -0,0 +1,287 @@
...
+use std::io::{BufReader, Take};
+
+#[derive(Debug)]
+pub enum Response {
+    /// Download action is not performed and no further action is required.
+    Skip(),
+    /// File was not downloaded, but hash action must be executed.
+    HashRequest(HashActionOptions),
+    /// Chunks of data to be downloaded.
+    DownloadData(Chunks<BufReader<Take<File>>>),
```
panhania, line 19: I would go with "collect" nomenclature rather than "download" (because technically what the agent does is actually "uploading"...).

src/action/finder/chunks.rs
Spawek: Done!

src/action/finder/chunks.rs
Spawek: Done! I think the code looks better after this change.

src/action/finder/chunks.rs
Spawek: Added.

src/action/finder/download.rs
Spawek: Changed to `CollectData`. Changed the comment to `/// Chunks of data to be uploaded.`.

src/action/finder/chunks.rs
Spawek: Fixed.

src/action/finder/chunks.rs
Spawek: Added.

src/action/finder/chunks.rs
Spawek: I formatted the whole file.

src/action/finder/chunks.rs
Spawek: Changed to AsRef<Path>.

src/action/finder/chunks.rs
```diff
@@ -0,0 +1,180 @@
...
+        Ok(file) => file,
+        Err(err) => {
+            warn!(
+                "failed to open file: {}, error: {}",
+                path.as_ref().display(),
+                err
+            );
+            return None;
+        }
+    };
```
panhania, line 60: This can be simplified with `ack! { ... }?;`.

src/action/finder/chunks.rs
```diff
@@ -0,0 +1,180 @@
...
+
+    if let Err(err) = file.seek(SeekFrom::Start(offset)) {
+        warn!(
+            "failed to seek in file: {}, error: {}",
+            path.as_ref().display(),
+            err
+        );
+        return None;
+    }
```
panhania, line 69: Ditto.

src/action/finder/download.rs
```diff
@@ -0,0 +1,295 @@
...
+pub struct ChunkId {
+    /// A SHA-256 digest of the referenced chunk data.
+    sha256: [u8; 32],
+    offset: u64,
+    length: u64,
+}
+
+impl ChunkId {
+    /// Creates a chunk identifier for the given chunk.
+    pub fn make(chunk: &Vec<u8>, offset: u64) -> ChunkId {
```
panhania, line 71: Prefer to use slices rather than references to vectors, it is almost always better.

src/action/finder/download.rs
```diff
@@ -0,0 +1,295 @@
...
+        assert_eq!(
+            chunks.next().unwrap().unwrap(),
+            "e_".bytes().collect::<Vec<_>>()
+        );
+        assert!(chunks.next().is_none());
+    }
+
+    #[test]
+    fn test_chunk_id() {
+        let chunk = ChunkId::make(&"some_test_data".as_bytes().to_vec(), 5);
```
panhania, line 282: If you apply the suggestion with using slices, then you can use bytestring literal here.

panhania (changes requested):

src/action/finder/chunks.rs
Spawek: Done! Looks much better now.

src/action/finder/chunks.rs
Spawek: Changed! Looks much better too.

src/action/finder/download.rs
Spawek: Changed to slice.

src/action/finder/download.rs
Spawek: Done!

panhania (approved):

PR #52, opened 2021-03-05
Add support for conditions in CSFF.

Spawek:
Exactly the same code for `chunks.rs` is in "download" Pull Request - please review it in there first.


panhania (approved):
There are still some minor issues, but I guess they can be fleshed out at some later point.

PR #53, opened 2022-11-01
Replace the `rust-crypto` crate

diceride:
Replaces the [rust-crypto](https://crates.io/crates/rust-crypto) crate with distributed implementations of various hash algorithms that use the traits from the [digest](https://crates.io/crates/digest) crate. See [RustCrypto/hashes](https://github.com/RustCrypto/hashes) for implementations which use this trait.

* SHA-1 hash: [sha1](https://crates.io/crates/sha1)
* SHA-256 hash: [sha2](https://crates.io/crates/sha2)
* MD5 hash: [md-5](https://crates.io/crates/md-5)

### DIFF
--- a/src/action/finder/hash.rs
+++ b/src/action/finder/hash.rs
@@ -1,8 +1,10 @@
 use crate::action::finder::request::HashActionOptions;
 use crate::fs::Entry;
-use crypto::digest::Digest;
 use log::warn;
+use md5::Md5;
 use rrg_macro::ack;
+use sha1::Sha1;
+use sha2::{Digest, Sha256};
 use std::cmp::min;
 use std::fs::File;
 use std::io::Read;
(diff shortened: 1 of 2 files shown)


src/action/finder/hash.rs
```diff
@@ -1,40 +1,42 @@
 use crate::action::finder::request::HashActionOptions;
 use crate::fs::Entry;
-use crypto::digest::Digest;
 use log::warn;
+use md5::Md5;
 use rrg_macro::ack;
+use sha1::Sha1;
+use sha2::{Digest, Sha256};
```
panhania, line 7: Since `Digest` is shared here with `Md5`, `Sha1` and `Sha256`, I think it makes sense to state the dependency on the [`digest`] crate explicitly and import `Digest` from there.

Also, for trait improts that are brough into the scope only for the extension methods, prefer to use the [`use Foo as _;` syntax][1].

[`digest`]: https://crates.io/crates/digest

[1]: https://doc.rust-lang.org/reference/items/use-declarations.html#underscore-imports

panhania (changes requested):
Thank you for the pull request! Left a small comment, but otherwise looks good!

PR #54, opened 2022-11-01
Add the GetHostname action

### LINKED ISSUE #4: Implement the `GetHostname` action
RRG should support the [`GetHostname`][hostname-action] action. It does not have any input arguments and should respond with [`DataBlob`][hostname-response].

[hostname-action]: https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/client/grr_response_client/client_actions/admin.py#L50-L57
[hostname-response]:
https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/proto/grr_response_proto/jobs.proto#L401-L423

diceride:
Adds the [`GetHostname`](https://github.com/google/grr/blob/f0c26b1e83e77a8f53f7b7aef706b418061682ae/grr/client/grr_response_client/client_actions/admin.py#L50-L57) action. The implementation follows Python's [`socket.gethostname`](https://docs.python.org/3/library/socket.html#socket.gethostname). This PR adds Microsoft's [windows](https://crates.io/crates/windows) crate that provides convenient bindings for Windows.

* unix: [syscall gethostname](https://linux.die.net/man/2/gethostname) ([`libc::gethostname`](https://docs.rs/libc/0.2.137/libc/fn.gethostname.html))
* windows: [GetComputerNameExW function (sysinfoapi.h)](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getcomputernameexw) using the [windows](https://crates.io/crates/windows) crate.

Tests

The test is inspired by Go's [`os.Hostname`](https://pkg.go.dev/os#Hostname) test and uses the hostname program. The test will be skipped, if the hostname program does not exist. On Linux, the hostname program is not installed on all distributions by default (the program exists for example on Ubuntu).

* linux: https://linux.die.net/man/1/hostname
* bsd (darwin): https://www.freebsd.org/cgi/man.cgi?query=hostname
* windows: https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/hostname

References

* Python socketmodule.c https://github.com/python/cpython/blob/8b24d60f1b7725d526ea5b5d3904b72e8b41ae0d/Modules/socketmodule.c#L5569
* sys_linux.go https://github.com/golang/go/blob/f229e7031a6efb2f23241b5da000c3b3203081d6/src/os/sys_linux.go#L12
* sys_windows.go https://github.com/golang/go/blob/f229e7031a6efb2f23241b5da000c3b3203081d6/src/os/sys_windows.go
* sys_bsd.go (darwin) https://github.com/golang/go/blob/f229e7031a6efb2f23241b5da000c3b3203081d6/src/os/sys_bsd.go

Resolves #4

### DIFF
--- a/src/action/hostname.rs
+++ b/src/action/hostname.rs
@@ -10,24 +10,27 @@ use libc::c_char;
 use log::error;
 #[cfg(target_family = "unix")]
 use std::ffi::CStr;
-#[cfg(target_os = "windows")]
 use std::ffi::OsString;
+#[cfg(target_family = "unix")]
+use std::io::Error;
+#[cfg(target_family = "unix")]
+use std::os::unix::prelude::OsStringExt as _;
 #[cfg(target_os = "windows")]
-use std::os::windows::ffi::OsStringExt;
+use std::os::windows::ffi::OsStringExt as _;
 #[cfg(target_os = "windows")]
 use windows::{
+    core::{Error, PWSTR},
     Win32::System::SystemInformation::{
         ComputerNamePhysicalDnsHostname, GetComputerNameExW,
     },
-    core::PWSTR
 };
 use crate::session::{self, Session};
 
 /// A response type for the hostname action.
 struct Response {
-    /// The host name reported by the kernel, or `None` if the attemps to
+    /// The host name reported by the kernel, or an error if the attemps to
     /// obtain the host name failed.
-    hostname: Option<String>,
+    hostname: Result<OsString, Error>,
 }
 
 impl super::Item for Response {
@@ -37,15 +40,15 @@ impl super::Item for Response {
 
     fn into_proto(self) -> Self::Proto {
         let hostname = match self.hostname {
-            Some(hostname) => hostname,
-            None => {
-                error!("cannot get hostname, all methods failed");
-                String::new()
+            Ok(hostname) => hostname,
+            Err(err) => {
+                error!("{}", err);
+                OsString::new()
             },
         };
 
         let mut proto = protobuf::well_known_types::StringValue::new();
-        proto.set_value(hostname);
+        proto.set_value(hostname.to_string_lossy().to_string());
 
         proto
     }
@@ -53,34 +56,35 @@ impl super::Item for Response {
 
 /// Obtains system hostname (Unix version).
 ///
-/// This function returns `None` in case of errors.
+/// This function returns `std::io::Error` in case of errors.
 #[cfg(target_family = "unix")]
-fn get_hostname() -> Option<String> {
+fn get_hostname() -> Result<OsString, Error> {
 	let size =
         unsafe { libc::sysconf(libc::_SC_HOST_NAME_MAX) as libc::size_t };
 
-    let mut buf = vec![0 as c_char; size];
+    let mut buf = vec![0_u8 as c_char; size + 1];
 
-    let p = buf.as_mut_ptr();
     let result = unsafe {
-        libc::gethostname(p, size)
+        libc::gethostname(buf.as_mut_ptr(), size)
     };
 
     if result != 0 {
-        None
+        Err(Error::last_os_error())
     } else {
-        let hostname = 
-            unsafe { String::from_utf8_lossy(CStr::from_ptr(p).to_bytes()).into() };
+        buf.push(0);
+        let hostname = unsafe {
+            OsString::from_vec(CStr::from_ptr(buf.as_ptr()).to_bytes().to_vec())
+        };
 
-        Some(hostname)
+        Ok(hostname)
     }
 }
 
 /// Obtains system hostname (Windows version).
 ///
-/// This function returns `None` in case of errors.
+/// This function returns `windows::core::Error` in case of errors.
 #[cfg(target_os = "windows")]
-fn get_hostname() -> Option<String> {
+fn get_hostname() -> Result<OsString, Error> {
     let mut size = 0;
     unsafe {
         GetComputerNameExW(
@@ -99,19 +103,19 @@ fn get_hostname() -> Option<String> {
         )
     };
 
-    if result.as_bool() {
-        unsafe { buf.set_len(size as usize); }
+    match result.ok() {
+        Ok(()) => {
+            unsafe { buf.set_len(size as usize); }
 
-        Some(OsString::from_wide(&buf).to_string_lossy().to_string())
-    } else {
-        None
+            Ok(OsString::from_wide(&buf))
+        },
+        Err(err) => Err(err)
     }
 }
 
 /// Handles requests for the hostname action.
 pub fn handle<S: Session>(session: &mut S, _: ()) -> session::Result<()> {
-    session.reply(Response {hostname: get_hostname()})?;
-    Ok(())
+    session.reply(Response {hostname: get_hostname()})
 }
 
 #[cfg(test)]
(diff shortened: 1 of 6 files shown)


panhania (changes requested):

src/action/hostname.rs
```diff
@@ -0,0 +1,151 @@
...
+
+    #[test]
+    fn test_hostname() {
+        let output = match Command::new("hostname")
+            .stdout(Stdio::piped())
+            .spawn() {
+            Ok(child) => child.wait_with_output().ok().unwrap(),
+            Err(err) => {
+                if let ErrorKind::NotFound = err.kind() {
+                    return
```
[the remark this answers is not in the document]
diceride: Ah, I did not notice this part in the Cargo.toml file. This is a very elegant solution, I like it.

src/action/hostname.rs
```diff
@@ -0,0 +1,151 @@
...
+    },
+    core::PWSTR
+};
+use crate::session::{self, Session};
+
+/// A response type for the hostname action.
+struct Response {
+    /// The host name reported by the kernel, or `None` if the attemps to
+    /// obtain the host name failed.
+    hostname: Option<String>,
```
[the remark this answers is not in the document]
diceride: Yes, I was not sure how to handle this. In the end, the OsString will be converted to String when transmitted, so it is potentially lossy. But the API should use OsString, I agree.

src/action/hostname.rs
```diff
@@ -0,0 +1,151 @@
...
+        ComputerNamePhysicalDnsHostname, GetComputerNameExW,
+    },
+    core::PWSTR
+};
+use crate::session::{self, Session};
+
+/// A response type for the hostname action.
+struct Response {
+    /// The host name reported by the kernel, or `None` if the attemps to
+    /// obtain the host name failed.
```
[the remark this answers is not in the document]
diceride: Yes, very good. While I was working on this, I was not sure and oriented myself on the existing code base.

src/action/hostname.rs
```diff
@@ -0,0 +1,151 @@
...
+fn get_hostname() -> Option<String> {
+	let size =
+        unsafe { libc::sysconf(libc::_SC_HOST_NAME_MAX) as libc::size_t };
+
+    let mut buf = vec![0 as c_char; size];
+
+    let p = buf.as_mut_ptr();
+    let result = unsafe {
+        libc::gethostname(p, size)
+    };
```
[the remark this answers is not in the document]
diceride: A very good catch. We are converting the buffer to a C string first, before converting it into a Rust string, thus dropping the null byte.

diceride:
The function returns a Result containing OsString or an platform-specific error representing the last OS error. On Windows, this is the calling thread's last-error code value.

windows::core::Error::from_win32: https://github.com/microsoft/windows-rs/blob/master/crates/libs/windows/src/core/error.rs#L29

https://learn.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-getlasterror

diceride:
The function returns a Result containing OsString or an platform-specific error representing the last OS error (thread-safe).

diceride:
While I was working on the platform info action, I realized that I made a mistake. The hostname action function throws an error now, I think this is what you have meant.

src/action/hostname.rs
```diff
@@ -0,0 +1,136 @@
...
+use std::io::Error;
+#[cfg(target_family = "unix")]
+use std::os::unix::prelude::OsStringExt as _;
+#[cfg(target_os = "windows")]
+use std::os::windows::ffi::OsStringExt as _;
+#[cfg(target_os = "windows")]
+use windows::{
+    core::{Error, PWSTR},
+    Win32::System::SystemInformation::{ComputerNamePhysicalDnsHostname, GetComputerNameExW},
+};
```
panhania, line 24: You can scope imports under the functions that actually use them. Otherwise there is a bit of a mess, e.g. `Error` refers to completely different types depending on the platform (`std::io::Error` on Linux, `std::windows::core::Error` on Windows). Sometimes it is useful but I think it just increases confusion here.

src/action/hostname.rs
```diff
@@ -0,0 +1,136 @@
...
+    use std::io::{BufRead, ErrorKind};
+    use std::process::{Command, Stdio};
+
+    #[test]
+    fn test_hostname() {
+        let mut session = session::FakeSession::new();
+
+        if let Err(err) = handle(&mut session, ()) {
+            panic!("{:?}", err);
+        };
```
panhania, line 117: Pretty much equivalent to `handle(&mut session, ()).unwrap()`.

src/action/hostname.rs
```diff
@@ -0,0 +1,136 @@
...
+
+        assert_eq!(session.reply_count(), 1);
+        let response: &Response = session.reply(0);
+        let hostname = response.hostname.as_ref().unwrap();
+
+        let output = match Command::new("hostname").stdout(Stdio::piped()).spawn() {
+            Ok(child) => child.wait_with_output().ok().unwrap(),
+            Err(err) if err.kind() == ErrorKind::NotFound => return,
```
panhania, line 125: Since we use the `test-hostname` feature now, I think this branch can (and even should to ensure that the test works as expected) be removed.

src/action/hostname.rs
```diff
@@ -0,0 +1,136 @@
...
+            Err(err) => panic!("{}", err),
```
panhania, line 126: Similarly, this can be simplified with an `.unwrap()`.

src/session/error.rs
```diff
@@ -19,6 +19,12 @@ pub struct Error {
 pub enum ErrorKind {
     /// The action execution failed.
     ExecutionFailure,
+    /// I/O error.
+    #[cfg(target_family = "unix")]
+    IoError(std::io::ErrorKind),
+    /// Windows error.
+    #[cfg(target_os = "windows")]
+    WindowsError,
```
panhania, line 27: For me these are subcases of `ExecutionFailure` and thus should be defined as errors for individual actions. Take a look how the `listdir` action [does it][1].

[1]: https://github.com/google/rrg/blob/a7eb4a0a2f9c3818eef760e96d562100868d5fc5/src/action/listdir.rs#L36-L72

panhania (changes requested):

src/action/hostname.rs
diceride: Yes, I was not sure what the general coding style is. In general, I prefer to scope imports too.

src/session/error.rs
diceride: I created a new error type. The signature of the `get_hostname` function now throws `std::io::Error` for all platforms

diceride:
- Updated the error and the `get_hostname` function signature. It returns `std::io::Error` for all platforms now.
- Replaced the windows crate with the lightweight windows-sys crate.
- Windows: I found the [`MAX_COMPUTERNAME_LENGTH`](https://docs.rs/windows-sys/latest/windows_sys/Win32/System/WindowsProgramming/constant.MAX_COMPUTERNAME_LENGTH.html) const and used it for the MaybeUninit wrapper type to construct the buffer.

panhania (approved):

panhania:
Same comment as in #56: looks good but there are merge conflicts (and are still going to be be after we merge #56, so maybe it makes sense to wait).

diceride:
Yes, please merge #56 first and then I am going to rebase this shebang again.

diceride:
Yes, please merge #56 first and then I am going to rebase this shebang again. And once it is merged, I am going to continue to work on #55.

diceride:
I rebased from upstream master.

diceride:
There is a flaky network test on MacOS (unrelated to this PR).

diceride:
I should have merged the master branch instead of rebasing, but I stopped caring about git workflows years ago. Code is just garbage.

panhania:
@diceride: This looks good to be merged, are you sure you wanted to close this PR?

PR #55, opened 2022-11-06
Add the GetPlatformInfo action

### LINKED ISSUE #6: Implement the `GetPlatformInfo` action

diceride:
Adds the [GetPlatformInfo](https://github.com/google/grr/blob/ce2b70f9540ee17d127cd8cdc79a04a3f4e2ce73/grr/client/grr_response_client/client_actions/admin.py#L60-L66) action.

The implementation is following the Python cpython runtime, but is not as backwards compatible.

Known compatibility requirements

- MacOS: >=10.10
- Windows: >=7 (?)

Comparison matrix table of the protocol buffer properties ([source code](https://github.com/google/grr/blob/ce2b70f9540ee17d127cd8cdc79a04a3f4e2ce73/grr/proto/grr_response_proto/jobs.proto#L892))

| Property | grr | rrg | grr | rrg | grr | rrg |
|---|---|---|---|---|---|---|
|  | Linux | Linux | MacOS | MacOS | Windows | Windows |
| `system` | Yes | Yes | Yes | Yes | Yes | Yes (*) |
| `node` | Yes | Yes | Yes | Yes | Yes | Yes |
| `release` | Yes | Yes (*) | Yes | Yes | Yes | Yes |
| `version` | Yes | Yes (*) | Yes | Yes (*) | Yes | Yes (*) |
| `machine` | Yes | Yes | Yes | Yes | Yes | Yes (*) |
| `kernel` | Yes | Yes | Yes | Yes | Yes | Yes |
| `fqdn` | Yes | Yes | Yes | Yes | Yes | Yes |
| `install_date` |   | Yes  |   | Yes  |   | Yes  |
| `libc_ver` |   | Yes (*) |   |   |   |   |
| `architecture` | Yes | Yes | Yes | Yes | Yes | Yes |
| `pep425tag` | Yes |  | Yes |  | Yes |  |

Differences (marked with an asterisk * in the comparison matrix table above)

- Linux: The grr client uses the Python [distro](https://pypi.org/project/distro) package to generate _human readable_ values for the `release` and `version` properties. As the library is quite extensive ([source code](https://github.com/python-distro/distro/blob/master/src/distro/distro.py)), the raw values from libc::utsname are used instead. The conversion to a human readable format should be done on the server.
- The `pep425tag` property is always unset, as it only exists on Python to identify the unique signature of a Python runtime ([PEP 425 specification](https://peps.python.org/pep-0425/)).
- Linux: The `libc_version` property is only set, if the client was compiled against the GNU libc library.
- MacOS: The `release` property is hardcoded and set to the value "OSX" by default, similar to the grr client ([source code](https://github.com/google/grr/blob/ce2b70f9540ee17d127cd8cdc79a04a3f4e2ce73/grr/core/grr_response_core/lib/rdfvalues/client.py#L555)).
- MacOS: The `version` property is set using the [NSProcessInfo.processInfo function](https://developer.apple.com/documentation/foundation/nsprocessinfo/1408734-processinfo). The interface containing the operating system version is available equal to or after MacOS 10.10. In the Python cpython runtime implementation, the MacOS version is read using the SystemVersion.plist file ([source code](https://github.com/python/cpython/blob/de33df27aaf930be6a34027c530a651f0b4c91f5/Lib/platform.py#L477)).
- Windows: The `system` property is hardcoded and set to the value "Windows" by default.
- Windows: The `version` property does not include the service pack version. The service pack version can be read using the [GetVersionExA function](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getversionexa), but this function might be altered or unavailable for releases after Windows 8.1. The concept of service packs is obsolete since Windows 10.
- Windows: The `machine` property is set using the [GetVersion function](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getversion) and uses the [SYSTEM_INFO structure](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/ns-sysinfoapi-system_info#members) to determine the processor architecture.

Example - Linux

```bash
PlatformInfo { system: "Linux", node: "fv-az371-529", release: "5.15.0-1022-azure", version: Some("#27~20.04.1-Ubuntu SMP Mon Oct 17 02:03:50 UTC 2022"), machine: Some("x86_64"), kernel: "5.15.0-1022-azure", fqdn: "fv-az371-529.ipdn00qvwfrujbrcphh0gwusye.dx.internal.cloudapp.net", install_date: Some(SystemTime { tv_sec: 1667763649, tv_nsec: 0 }), libc_ver: Some("2.31"), architecture: "x86_64" }
```

Example - MacOS

```bash
PlatformInfo { system: "Darwin", node: "Mac-1667763736626.local", release: "OSX", version: Some("12.6.1"), machine: Some("x86_64"), kernel: "21.6.0", fqdn: "mac-1667763736626.local", install_date: Some(SystemTime { tv_sec: 1667763649, tv_nsec: 480097701 }), libc_ver: None, architecture: "x86_64" }
```

Example - Windows

```bash
PlatformInfo { system: "Windows", node: "fv-az449-397", release: "10", version: Some("10.0 (20348)"), machine: Some("x86_64"), kernel: "10.0.20348 Build 20348", fqdn: "fv-az449-397", install_date: Some(SystemTime { intervals: 133117186730000000 }), libc_ver: None, architecture: "x86_64" }
```

References

- Unix: [uname](https://www.ibm.com/docs/en/zos/latest?topic=descriptions-uname)
- Unix: [getaddrinfo](https://www.ibm.com/docs/en/zos/latest?topic=functions-getaddrinfo-get-address-information)
- Linux: [gnu_get_libc_version](https://www.gnu.org/software/gnulib/manual/html_node/gnu_005fget_005flibc_005fversion.html)
- MacOS: [NSProcessInfo.processInfo function](https://developer.apple.com/documentation/foundation/nsprocessinfo/1408734-processinfo)
- Windows: [GetComputerNameExW function](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getcomputernameexw)
- Windows: [GetFileVersionInfoSizeW function](https://learn.microsoft.com/en-us/windows/win32/api/winver/nf-winver-getfileversioninfosizew)
- Windows: [GetModuleHandleW function](https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-getmodulehandlew)
- Windows: [GetNativeSystemInfo function](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/nf-sysinfoapi-getnativesysteminfo)
- Windows: [VerQueryValueW function](https://learn.microsoft.com/en-us/windows/win32/api/winver/nf-winver-verqueryvaluew)
- Windows: [SYSTEM_INFO structure](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/ns-sysinfoapi-system_info)

Resolves #6

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -45,12 +45,20 @@ proc-mounts = { version = "0.2.4", optional = true }
 # [1]: https://github.com/rust-lang/cargo/issues/1596
 fuse = { version = "0.3.1", optional = true }
 
+[target.'cfg(target_os = "macos")'.dependencies]
+cocoa = { git = "https://github.com/servo/core-foundation-rs.git", rev = "786895643140fa0ee4f913d7b4aeb0c4626b2085", optional = true }
+objc = { version = "0.2", optional = true }
+
 [target.'cfg(target_os = "windows")'.dependencies]
+windows = { version = "0.43.0", features = ["Win32_Foundation", "Win32_System_Diagnostics_Debug", "Win32_Storage_FileSystem", "Win32_System_LibraryLoader", "Win32_System_SystemInformation"], optional = true }
 winreg = { version = "0.7.0", optional = true }
 
 [dev-dependencies]
 rand = { version = "0.8.5" }
 tempfile = { version = "3.3.0" }
+
+[target.'cfg(target_os = "linux")'.build-dependencies]
+cc = { version = "1.0", optional = true }
 
 [features]
 default = [
@@ -62,6 +70,7 @@ default = [
     "action-memsize",
     "action-metadata",
     "action-network",
+    "action-platform-info",
     "action-stat",
     "action-timeline",
 ]
@@ -74,6 +83,7 @@ action-listdir = []
 action-memsize = ["dep:sysinfo"]
 action-metadata = []
 action-network = ["dep:netstat2", "dep:sysinfo"]
+action-platform-info = ["dep:cc", "dep:cocoa", "dep:objc", "dep:windows"]
 action-stat = []
 action-timeline = ["dep:flate2", "dep:sha2"]
 
--- a/build.rs
+++ b/build.rs
@@ -1,0 +1,10 @@
+fn main() {
+    #[cfg(all(target_os = "linux", feature = "action-platform-info"))]
+    {
+        println!("cargo:rerun-if-changed=src/action/libc_version.c");
+
+        cc::Build::new()
+            .file("src/action/libc_version.c")
+            .compile("liblibc_version.a");
+    }
+}
--- a/src/action/libc_version.c
+++ b/src/action/libc_version.c
@@ -1,0 +1,14 @@
+#include <features.h>
+#ifdef __GNU_LIBRARY__
+#include <gnu/libc-version.h>
+#else
+#include <stddef.h>
+#endif
+
+const char *libc_version(void) {
+#ifdef __GNU_LIBRARY__
+  return gnu_get_libc_version();
+#else
+  return NULL;
+#endif
+}
--- a/src/action/platform_info.rs
+++ b/src/action/platform_info.rs
@@ -1,0 +1,394 @@
+// Copyright 2020 Google LLC
+//
+// Use of this source code is governed by an MIT-style license that can be found
+// in the LICENSE file or at https://opensource.org/licenses/MIT.
+
+//! A handler and associated types for the platform info action.
+
+use crate::action::insttime::get_install_time;
+#[cfg(target_os = "windows")]
+use crate::fs::windows::dll_module_fileinfo;
+use crate::session::{self, Session};
+#[cfg(target_os = "macos")]
+use cocoa::{
+    appkit::*,
+    base::nil,
+    foundation::{NSInteger, NSProcessInfo},
+};
+use log::error;
+#[cfg(target_os = "macos")]
+use objc::{msg_send, sel, sel_impl};
+use std::env::consts::ARCH;
+#[cfg(target_family = "unix")]
+use std::ffi::CStr;
+use std::ffi::OsString;
+#[cfg(target_family = "unix")]
+use std::io::Error;
+#[cfg(target_family = "unix")]
+use std::mem;
+#[cfg(target_family = "unix")]
+use std::os::unix::prelude::OsStringExt as _;
+#[cfg(target_os = "windows")]
+use std::os::windows::ffi::OsStringExt as _;
+#[cfg(target_family = "unix")]
+use std::ptr;
+use std::time::{SystemTime, UNIX_EPOCH};
+#[cfg(target_os = "windows")]
+use windows::{
+    core::{Error, PWSTR},
+    w,
+    Win32::System::{
+        Diagnostics::Debug::{
+            PROCESSOR_ARCHITECTURE, PROCESSOR_ARCHITECTURE_AMD64, PROCESSOR_ARCHITECTURE_ARM,
+            PROCESSOR_ARCHITECTURE_IA64, PROCESSOR_ARCHITECTURE_INTEL,
+        },
+        SystemInformation::{
+            ComputerNamePhysicalDnsFullyQualified, ComputerNamePhysicalDnsHostname,
+            GetComputerNameExW, GetNativeSystemInfo, SYSTEM_INFO,
+        },
+    },
+};
+
+#[cfg(target_os = "linux")]
+#[link(name = "libc_version")]
+extern "C" {
+    fn libc_version() -> *const libc::c_char;
+}
+
+#[cfg(target_os = "macos")]
+#[allow(dead_code)]
+#[repr(C)]
+struct NSOperatingSystemVersion {
+    major_version: NSInteger,
+    minor_version: NSInteger,
+    patch_version: NSInteger,
+}
+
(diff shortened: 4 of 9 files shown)


diceride:
@panhania  I would to like to know, whether the provided platform information is sufficient, before splitting up the code up into smaller functions and adding tests (eg. using the Windows [ver](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/ver) command). 

For example, the `release` property value for MacOS is fixed. On Windows, I had trouble to determine the correct build version, as it is reported by the System Information app.

diceride:
@panhania Please let me know, whether the provided platform information is sufficient. In the next step, I am going to split up the code into smaller functions and add tests (eg. using the Windows [ver](https://learn.microsoft.com/en-us/windows-server/administration/windows-commands/ver) command).

build.rs
```diff
@@ -0,0 +1,10 @@
+fn main() {
+    #[cfg(all(target_os = "linux", feature = "action-platform-info"))]
+    {
+        println!("cargo:rerun-if-changed=src/action/libc_version.c");
+
+        cc::Build::new()
+            .file("src/action/libc_version.c")
+            .compile("liblibc_version.a");
+    }
+}
```
panhania, line 10: I am affraid build script is not an option — this does not play well with our internal build infrastructure (for `rrg-proto` I already have to do some awkward workarounds).

Cargo.toml
```diff
@@ -45,13 +45,21 @@ proc-mounts = { version = "0.2.4", optional = true }
 # [1]: https://github.com/rust-lang/cargo/issues/1596
 fuse = { version = "0.3.1", optional = true }
 
+[target.'cfg(target_os = "macos")'.dependencies]
+cocoa = { git = "https://github.com/servo/core-foundation-rs.git", rev = "786895643140fa0ee4f913d7b4aeb0c4626b2085", optional = true }
+objc = { version = "0.2", optional = true }
```
panhania, line 50: These seem like a quite heavy dependencies, I wonder if there is a way to avoid these. Moreover, I would prefer not to depend on anything that is not on crates.io because of our internal build infrastructure.

src/action/libc_version.c
```diff
@@ -0,0 +1,14 @@
...
+#include <stddef.h>
+#endif
+
+const char *libc_version(void) {
+#ifdef __GNU_LIBRARY__
+  return gnu_get_libc_version();
+#else
+  return NULL;
+#endif
+}
```
panhania, line 14: What is the benefit over [`libc::gnu_get_libc_version`]?

[`libc::gnu_get_libc_version`]: https://docs.rs/libc/0.2.137/libc/fn.gnu_get_libc_version.html

src/action/platform_info.rs
```diff
@@ -0,0 +1,394 @@
...
+struct Response {
+    /// The platform information.
+    platform_info: PlatformInfo,
+}
+
+/// The platform information.
+#[derive(Clone, Debug)]
+struct PlatformInfo {
+    /// The system platform (Windows|Darwin|Linux).
+    system: String,
```
panhania, line 77: Why not an `enum`?

src/action/platform_info.rs
```diff
@@ -0,0 +1,394 @@
...
+    /// The system's fully qualified domain name.
+    fqdn: OsString,
+    /// When system was installed.
+    install_date: Option<SystemTime>,
+    /// The C library version.
+    libc_ver: Option<String>,
+    /// The architecture of this binary. (Note this can be different from
+    /// the machine architecture in the case of a 32 bit binary running
+    /// on a 64 bit system)
+    architecture: String,
```
panhania, line 97: Since this just uses `std::env::consts::ARCH` this can be typed as `&'static str`.

panhania (commented):
Thanks for this pull request, a lot of solid research! I left some initial comments that can shape how the overall code looks in the future, so not diving into details yet.

src/action/libc_version.c
diceride: Yes, excellent. I did not know that the gnu_get_* APIs were added recently

https://github.com/rust-lang/libc/commit/f4150780462d20d709369743bbc7f7a9e26cb6ca

build.rs
diceride: Yes, we can scrap the build script now that the [libc::gnu_get_libc_version](https://docs.rs/libc/0.2.137/libc/fn.gnu_get_libc_version.html) function is available in the standard library.

diceride: Yes, we can scrap the build script. The [libc::gnu_get_libc_version](https://docs.rs/libc/0.2.137/libc/fn.gnu_get_libc_version.html) function is now available in the standard library.

PR #56, opened 2022-11-07
Remove dependency on the `winreg` crate

diceride:
Uses the [windows](https://crates.io/crates/windows) crate to search for the installation date in the registry. The aim is to limit the used Windows API crates. 

References
- Windows: [RegGetValueW function](https://learn.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-reggetvaluew)

### DIFF
--- a/src/action/insttime.rs
+++ b/src/action/insttime.rs
@@ -242,7 +242,8 @@ fn get_install_time() -> Option<SystemTime> {
 fn get_install_time() -> Option<SystemTime> {
     use std::ffi::c_void;
     use std::mem::{size_of, MaybeUninit};
-    use windows::{
+    use std::ptr::null_mut;
+    use windows_sys::{
         w,
         Win32::{
             Foundation::NO_ERROR,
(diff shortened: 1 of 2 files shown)


src/action/insttime.rs
```diff
@@ -240,14 +240,39 @@ fn get_install_time() -> Option<SystemTime> {
...
+            Some(install_date_data.as_mut_ptr() as *mut c_void),
+            Some(&mut install_date_size),
+        )
+    };
+
+    if result == NO_ERROR {
+        let install_date = unsafe { install_date_data.assume_init() };
+
+        if install_date > 0 {
+            return Some(SystemTime::UNIX_EPOCH + Duration::from_secs(install_date as u64))
```
panhania, line 271: Prefer not to use `as` casts as they don't shield from potential overflow issues. `u64::from` does not have these issues.

panhania (changes requested):

diceride:
- Replaced the windows crate with the lightweight windows-sys crate.
- Used `u64::from` to convert from u32 losslessly.

panhania (approved):

panhania:
@diceride Looks good now but has merge conflicts. Could you look into resolving them?

diceride:
Bim bam bum late night merge errors and broken Windows vm's are much fun 🐼

PR #62, opened 2024-06-24
Implement the `get_winreg_value` action

panhania:


proto/rrg/winreg.proto
```diff
@@ -0,0 +1,59 @@
...
+// [1]: https://learn.microsoft.com/en-us/windows/win32/sysinfo/predefined-keys
+enum PredefinedKey {
+    UNKNOWN = 0;
+    CLASSES_ROOT = 1;
+    CURRENT_USER = 2;
+    LOCAL_MACHINE = 3;
+    USERS = 4;
+    PERFORMANCE_DATA = 5;
+    CURRENT_CONFIG = 6;
+    PERFORMANCE_TEXT = 8;
```
s-westphal, line 24: Is the 7 missing here on purpose?

proto/rrg/winreg.proto
```diff
@@ -0,0 +1,59 @@
+// Copyright 2023 Google LLC
```
s-westphal, line 1: Should it be Copyright **2024**?

s-westphal (approved):

proto/rrg/winreg.proto
panhania: Ah, right. Fixed here and in `get_winreg_value.proto` and in `get_winreg_value.rs`.

proto/rrg/winreg.proto
panhania: Sort of no (but not fully). Initially there was one more value (`DYN_DATA`) but it is no longer supported in Windows NT and the `winreg` crate won't support it anyway, so I deleted it. I considered to put `reserved` for its value since the initial idea was to use the same values as `HKEY_*` constants from the `winreg.h` header but that turned out to be infeasible (out of range for Protocol Buffers). Since the constants are sequential and start at 0x80000000, one could still argue for keeping some 1:1 mapping (it would be enough to add 0x80000000) which is why I skipped 7, but then there is `UNKNOWN` option anyway that makes the mapping even more questionable and make it is better to not attempt to replicate it in any form.

So, I re-assigned values sequentially.

PR #68, opened 2024-08-09
Add support for limiting depth of `walk_dir`

panhania:


crates/rrg/src/fs.rs
```diff
@@ -81,48 +68,14 @@ pub fn walk_dir<P: AsRef<Path>>(root: P) -> std::io::Result<WalkDir> {
...
-/// ```
-pub fn list_dir<P: AsRef<Path>>(path: P) -> std::io::Result<ListDir> {
-    let iter = std::fs::read_dir(path)?;
-
-    Ok(ListDir {
-        iter: iter,
-    })
-}
-
 /// Iterator over entries in all subdirectories.
```
s-westphal, line 79: add "until `max_depth`"?

crates/rrg/src/fs.rs
```diff
@@ -135,13 +88,39 @@ pub fn list_dir<P: AsRef<Path>>(path: P) -> std::io::Result<ListDir> {
...
+    /// assert!(!paths.contains(&"/usr/lib".into()));
+    /// ```
+    pub fn with_max_depth(mut self, max_depth: u32) -> WalkDir {
+        assert!(max_depth > 0);
+
+        self.max_depth = max_depth;
+        self
+    }
+
     #[cfg(target_family = "unix")]
```
s-westphal, line 124: What happens on macos?

crates/rrg/src/fs.rs
```diff
@@ -203,7 +189,21 @@ impl std::iter::Iterator for ListDir {
     type Item = std::io::Result<Entry>;
```
s-westphal, line 189: Is this needed?

s-westphal (approved):

crates/rrg/src/fs.rs
panhania: `max_depth` is a private field of the struct and docs are for public consumption, it should not leak implementation details.

However, I added a paragraph with reference to the `with_max_depth` method.

crates/rrg/src/fs.rs
panhania: `unix` option of [`target_family`] covers both macOS and Linux, as opposed to [`target_os`] that disambiguates between the two.

[`target_family`]: https://doc.rust-lang.org/reference/conditional-compilation.html#target_family
[`target_os`]: https://doc.rust-lang.org/reference/conditional-compilation.html#target_os

crates/rrg/src/fs.rs
panhania: I am not sure what you mean by "needed". `Item` [is an associated type][1] on the `Iterator` trait.

[1]: https://doc.rust-lang.org/std/iter/trait.Iterator.html#required-associated-types

s-westphal: Oh I see, thank you, I thought it is a [type alias](https://doc.rust-lang.org/reference/items/type-aliases.html) which is then not used.

s-westphal (approved):

crates/rrg/src/fs.rs
panhania: Ah, okay. Just so you know then: if you see `type ... = ...;` declaration in trait implementation block (`impl ... for ... { ... }`), it _has to_ be an associated type, traditional aliases are not allowed there.

PR #70, opened 2024-09-25
Implement the `grep_file_contents` action

panhania:


proto/rrg/action/grep_file_contents.proto
```diff
@@ -0,0 +1,38 @@
...
+//
+// Use of this source code is governed by an MIT-style license that can be found
+// in the LICENSE file or at https://opensource.org/licenses/MIT.
+syntax = "proto3";
+
+package rrg.action.get_file_contents;
+
+import "rrg/fs.proto";
+
+message Args {
```
s-westphal, line 11: Would it make sense to add an option to return the whole line with the match, or also lines before and after, like the `-A` `-B` `-C`. (Might be also just something to consider for the future)

s-westphal (approved):

proto/rrg/action/grep_file_contents.proto
panhania: > Would it make sense to add an option to return the whole line with the match (...)

This can be easily achieved by rewriting the regex: instead of `foo` you match against `^.*foo.*$`, thus seems like a redundant functionality.

> (...) or also lines before and after, like the `-A` `-B` `-C`. (Might be also just something to consider for the future)

This is an interesting proposal that I can see being occasionally useful. I am hesitant to implement it right now as we don't have use case for it yet and it will interact in weirdly once we start enforcing line limit. There are two reasons for the limit: keeping agent memory usage in check and not going over the Fleetspeak message size limit. If we allow only up to 1 MiB of line data, we guarantee that we will not go over the Fleetspeak limit. With multiple lines this won't be the case—not impossible to make it work in non-pathological cases but it adds a lot of complexity.

But we can keep it in mind for the future.

PR #80, opened 2024-12-18
Add CLI argument for signature verification key

s-westphal:


crates/rrg/Cargo.toml
```diff
@@ -130,3 +130,12 @@ features = [
     "Win32_Foundation",
     "Win32_Storage_FileSystem",
 ]
+
+[dependencies.ed25519-dalek]
+version = "2.1.1"
+features = [
+    "rand_core",
+]
```
panhania, line 138: If I understand correctly, we only need `rand_core` for testing, so it should be only specified in `dev-dependencies`.

crates/rrg/src/args.rs
```diff
@@ -49,6 +49,14 @@ pub struct Args {
            arg_name="PATH",
            description="whether to log to a file")]
     pub log_to_file: Option<std::path::PathBuf>,
+
+    /// The public key for verfying signed commands.
+    #[argh(option,
+       long="command-verification-key",
+       arg_name="KEY",
+       description="verification key for signed commands",
```
panhania, line 57: Let's specify that it expects hex encoded bytes, e.g. _"verification key for signed commands (hex-encoded bytes)"_.

crates/rrg/Cargo.toml
```diff
@@ -130,3 +130,12 @@ features = [
...
+
+[dependencies.hex]
+version = "0.4.3"
```
panhania, line 141: Let's not bring a third-party dependency for something that can be trivially implemented in a few lines ourselves.

panhania (changes requested):

crates/rrg/Cargo.toml
s-westphal: Yes exactly! Updated.

crates/rrg/src/args.rs
s-westphal: Done!

crates/rrg/Cargo.toml
s-westphal: Alright, I added the function with some tests.

crates/rrg/src/args.rs
```diff
@@ -75,9 +85,58 @@ fn parse_duration(value: &str) -> Result<Duration, String> {
     humantime::parse_duration(value).map_err(|error| error.to_string())
 }
 
+/// Decodes a Vector of hex digits to a Vector of byte values.
+fn hex_decode(hex: Vec<u8>) -> Result<Vec<u8>, HexDecodeError> {
```
panhania, line 89: I understand that you named it after `hex::decode` but the convention for functions is that they usually start with a verb, so a more appropriate name would be `decode_hex` (and changing the function name should also affect name of the error struct).

panhania, line 89: Accepting an owned vector that is consumed by the function is indication that you should probably accept a slice (`&[u8]`).

The original [`hex::decode`](https://docs.rs/hex/latest/hex/fn.decode.html) goes even a step further an accepts `AsRef<[u8]>`, so will work both with `Vec`s and slices without any additional borrows, although I don't think we need that level of generality here.

crates/rrg/src/args.rs
```diff
@@ -59,6 +59,16 @@ pub struct Args {
     pub command_verification_key: Option<ed25519_dalek::VerifyingKey>,
 }
 
+#[derive(Debug)]
+struct HexDecodeError();
```
panhania, line 63: No need for `()`, you can just have `struct Foo;` (you will also need to change creation not to use parens).

crates/rrg/src/args.rs
```diff
@@ -75,9 +85,58 @@ fn parse_duration(value: &str) -> Result<Duration, String> {
...
+    fn test_invalid_char() {
+        let hex = vec![b'x', b'y'];
+        assert!(hex_decode(hex).is_err());
+    }
+
+    #[test]
+    pub fn test_empty() {
+        let empty: Vec<u8> = vec![];
+        assert_eq!(hex_decode(empty).unwrap(), vec![]);
+    }
```
panhania, line 141: No need to use `test_*` prefix in test names, `#[test]` annotation is enough and adding it just introduces noise. You can still occasionally see old tests with `test_*` prefix in the codebase but consider this legacy back from days when I was still young.

At the same time, tests *should* be prefixed with the name of the function or type they test. At the moment in this module we only test `hex_decode`, but the moment we want to test other stuff, things can start to get confusing.

To sum up, I would rename these (after including the function name change) like `decode_hex_invalid_length`, etc.

crates/rrg/src/args.rs
```diff
@@ -75,9 +85,58 @@ fn parse_duration(value: &str) -> Result<Duration, String> {
...
+
+    #[test]
+    fn test_invalid_char() {
+        let hex = vec![b'x', b'y'];
+        assert!(hex_decode(hex).is_err());
+    }
+
+    #[test]
+    pub fn test_empty() {
```
panhania, line 138: `pub` is unnecessary here.

crates/rrg/src/args.rs
```diff
@@ -75,9 +85,58 @@ fn parse_duration(value: &str) -> Result<Duration, String> {
...
+    ed25519_dalek::VerifyingKey::try_from(&bytes[..]).map_err(|error| error.to_string())
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn test_hex_decode() {
+        let hex = vec![b'a', b'2', b'8', b'F'];
```
panhania, line 121: If you apply the suggestion for the function to take slices instead of `Vec`s, you can use bytestring literals, e.g. `b"a28F"`.

panhania, line 121: I would split this into two cases: one with lowercase letters and one with uppercase letters.

panhania (changes requested):

crates/rrg/src/args.rs
s-westphal: Done.

s-westphal: Done.

crates/rrg/src/args.rs
s-westphal: Done.

crates/rrg/src/args.rs
s-westphal: Done.

s-westphal: Done.

crates/rrg/src/args.rs
s-westphal: Done.

crates/rrg/src/args.rs
s-westphal: Done.

panhania (approved):

PR #82, opened 2024-12-18
Implement the `execute_signed_command` action

s-westphal:


crates/rrg/src/action/execute_signed_command.rs
```diff
@@ -0,0 +1,533 @@
...
+    };
+    let truncated_stdout = length_stdout > MAX_OUTPUT_SIZE;
+    if truncated_stdout {
+        stdout.truncate(MAX_OUTPUT_SIZE);
+    };
+
+    let mut stderr = Vec::<u8>::new();
+    let length_stderr = match command_process.stderr.take() {
+        Some(mut process_stderr) => process_stderr
+            .read_to_end(&mut stderr)
```
panhania, line 143: Ditto.

crates/rrg/src/action/execute_signed_command.rs
```diff
@@ -0,0 +1,533 @@
...
+        let item = session.reply::<Item>(0);
+
+        assert!(item.stderr.is_empty());
+        assert!(item.stdout.is_empty());
+        assert!(!item.exit_status.success());
+        #[cfg(target_family = "unix")]
+        {
+            use std::os::unix::process::ExitStatusExt;
+
+            assert_eq!(item.exit_status.signal(), Some(9)); // killed
```
panhania, line 530: I think `libc::SIGKILL` is more obvious.

crates/rrg/src/action/execute_signed_command.rs
```diff
@@ -0,0 +1,533 @@
...
+/// Handles invocations of the `execute_signed_command` action.
+pub fn handle<S>(session: &mut S, mut args: Args) -> crate::session::Result<()>
+where
+    S: crate::session::Session,
+{
+    match session.args().command_verification_key {
+        Some(key) => key
+            .verify_strict(&args.raw_command, &args.ed25519_signature)
+            .map_err(crate::session::Error::action)?,
+        None => panic!("missing verification key for signed command"),
```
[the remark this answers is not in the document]
s-westphal: Done.

crates/rrg/src/action/execute_signed_command.rs
```diff
@@ -0,0 +1,533 @@
...
+        .stdout(std::process::Stdio::piped())
+        .stderr(std::process::Stdio::piped())
+        .spawn()
+        .map_err(crate::session::Error::action)?;
+
+    let command_start_time = std::time::SystemTime::now();
+
+    let mut command_stdin = match command_process.stdin.take() {
+        Some(command_stdin) => command_stdin,
+        None => panic!("command stdin pipe should not be None"),
```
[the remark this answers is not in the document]
s-westphal: Done.

crates/rrg/src/action/execute_signed_command.rs
```diff
@@ -0,0 +1,533 @@
...
+    };
+    let handle = std::thread::spawn(move || {
+        match args.stdin {
+            Stdin::Signed(signed) => command_stdin
+                .write(&signed[..])
+                .expect("Failed to write to stdin"),
+            Stdin::Unsigned(unsigned) => command_stdin
+                .write(&unsigned[..])
+                .expect("Failed to write to stdin"),
```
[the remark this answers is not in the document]
s-westphal: This exception is in the spawned process, so I believe I cannot use `.map_error(...)`. The documentation says:
If the spawned thread panics, [join] will return an [Err] containing the argument given to [panic](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html). 
So that error should be returned in the next line, but it is of type `Box<dyn std::any::Any + Send>`  which I wasn't able to convert to a `create::session:Error`, see my comment in the next line.

crates/rrg/src/action/execute_signed_command.rs
```diff
@@ -0,0 +1,533 @@
...
+    });
+
+    // TODO: join returns a `Box<dyn std::any::Any + Send>`
+    // error which cannot be passed to crate:session:Error::action.
+    let _ = handle.join().unwrap();
+
+    while std::time::SystemTime::now()
+        .duration_since(command_start_time)
+        .unwrap()
+        < args.timeout
```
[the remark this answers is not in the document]
s-westphal: Nice, thank you!

crates/rrg/src/action/execute_signed_command.rs
```diff
@@ -0,0 +1,533 @@
...
+    };
+
+    let command_path = &std::path::PathBuf::try_from(args.command.take_path())
+        .map_err(crate::session::Error::action)?;
+
+    // For calling `cmd /C <command>` in windows, C:\Windows\System32
+    // needs to be on the PATH, otherwise <command> is not found.
+    let filtered_env: HashMap<String, String> = env::vars()
+        .filter(|&(ref k, _)| k.to_lowercase() == "path")
+        .collect();
```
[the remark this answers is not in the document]
s-westphal: If we don't have `C:\Windows\System32` on the path, then the `<command>` in  `cmd /C <command>` needs to have an absolute path, which is possible (e.g. https://github.com/google/rrg/pull/82/commits/92b777c1cbf18c0ac66096d77279e1de9620e90c). However, I was thinking as different hosts might have different paths to the \System32 folder (not sure how often this is actually the case though), the stored commands would not be very reusable.  
But I don't have a strong opinion on what the preferred solution here is.

crates/rrg/src/action/execute_signed_command.rs
```diff
@@ -0,0 +1,568 @@
...
+
+    let mut command_stdin = match command_process.stdin.take() {
+        Some(command_stdin) => command_stdin,
+        None => return Err(crate::session::Error::action(CommandStdinError)),
+    };
+
+    let handle = std::thread::spawn(move || match args.stdin {
+        Stdin::Signed(signed) => command_stdin
+            .write(&signed[..])
+            .expect("Failed to write to stdin"),
```
panhania, line 120: Nit: expectations should not be capitalized like sentences (similarly to explicit messages passed to `panic!`).

crates/rrg/src/action/execute_signed_command.rs
```diff
@@ -0,0 +1,568 @@
...
+        Stdin::Unsigned(unsigned) => command_stdin
+            .write(&unsigned[..])
+            .expect("Failed to write to stdin"),
```
panhania, line 123: Ditto.

crates/rrg/src/action/execute_signed_command.rs
```diff
@@ -0,0 +1,568 @@
+// Copyright 2025 Google LLC
+//
+// Use of this source code is governed by an MIT-style license that can be found
+// in the LICENSE file or at https://opensource.org/licenses/MIT.
+
+use std::{
+    collections::HashMap,
+    env,
```
panhania, line 8: Unused according to GitHub Actions.

crates/rrg/src/action/execute_signed_command.rs
```diff
@@ -0,0 +1,568 @@
...
+    };
+
+    let command_path = &std::path::PathBuf::try_from(args.command.take_path())
+        .map_err(crate::session::Error::action)?;
+
+    // // For calling `cmd /C <command>` in windows, C:\Windows\System32
+    // // needs to be on the PATH, otherwise <command> is not found.
+    // let filtered_env: HashMap<String, String> = env::vars()
+    //     .filter(|&(ref k, _)| k.to_lowercase() == "path")
+    //     .collect();
```
panhania, line 97: Let's remove this commented out part (and the `.envs(filtered_env)` part below).

crates/rrg/src/action/execute_signed_command.rs
```diff
@@ -0,0 +1,568 @@
...
+            .expect("Failed to write to stdin"),
+        Stdin::Unsigned(unsigned) => command_stdin
+            .write(&unsigned[..])
+            .expect("Failed to write to stdin"),
+        Stdin::None => 0,
+    });
+
+    // TODO: join returns a `Box<dyn std::any::Any + Send>`
+    // error which cannot be passed to crate:session:Error::action.
+    let _ = handle.join().unwrap();
```
panhania, line 129: If we cannot wrap the error (I will take a look at it eventually, so let's leave the TODO comment), we should still not crash the agent in case of failure—we can still fail the action but with less detailed error message.

crates/rrg/src/action/execute_signed_command.rs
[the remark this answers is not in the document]
panhania: Ah, makes sense. I left two minor nits about error messages and a comment about `.join().unwrap()` but panicking in the subprocess thread makes sense to me.

crates/rrg/src/action/execute_signed_command.rs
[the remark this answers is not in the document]
panhania: Yeah, it's an interesting issue. Still, for the time being, I would not do any magic workarounds and do the simplest thing. If this simplest thing turns out not to work well in practice, we will have to get back to the drawing board but I think we should be fine.

crates/rrg/src/action/execute_signed_command.rs
s-westphal: Done.

crates/rrg/src/action/execute_signed_command.rs
s-westphal: Done

crates/rrg/src/action/execute_signed_command.rs
s-westphal: Done.

crates/rrg/src/action/execute_signed_command.rs
s-westphal: Done.

panhania (approved):

crates/rrg/src/action/execute_signed_command.rs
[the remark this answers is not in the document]
panhania: This comment was not correct: stdin cannot be `None` as it is always set when constructing the `Command` object, the server invocation cannot change this. This is reverted to use a panic (through `.expect(...)`) in #86.

PR #83, opened 2025-01-15
Implement the `get_tcp_response` action

panhania:


crates/rrg/src/action/get_tcp_response.rs
```diff
@@ -0,0 +1,335 @@
+// Copyright 2023 Google LLC
```
s-westphal, line 1: nit: Copyright 2025

proto/rrg/action/get_tcp_response.proto
```diff
@@ -0,0 +1,32 @@
+// Copyright 2023 Google LLC
```
s-westphal, line 1: Nit: Copyright 2025

crates/rrg/src/action/get_tcp_response.rs
```diff
@@ -0,0 +1,335 @@
...
+        })
+    }
+}
+
+impl crate::response::Item for Item {
+
+    type Proto = rrg_proto::get_tcp_response::Result;
+
+    fn into_proto(self) -> Self::Proto {
+        let mut proto = rrg_proto::get_tcp_response::Result::new();
```
s-westphal, line 198: Would it make sense to add a field to the result proto that indicates if the response was larger than 1MiB and cut off?

s-westphal (approved):

crates/rrg/src/action/get_tcp_response.rs
panhania: Yeah, good question. On one hand it makes sense, on the other hand if we decide to send multiple responses in the future it feels a bit weird.

I added a TODO comment to consider this in the future but for now I will skip it not to complicate the implementation further since we have no need for this case at the moment anyway (we only need this action for cloud VM metadata collection which should be small HTTP response).

Issue #58, opened 2023-10-24
`ospect::os::fqdn` does not work on macOS


panhania:
Some change recently caused the tests for `ospect::os::fqdn` to fail. GitHub Actions run for 34441db ([run][1]) succeeded but the following 1a39af0 ([run][2]) failed. The commit does not touch this function at all.

[1]: https://github.com/google/rrg/actions/runs/6499161177
[2]: https://github.com/google/rrg/actions/runs/6499519330

When testing locally the low-level function used to implement this ([`getaddrinfo`]), the same error happens:

[`getaddrinfo`]: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/getaddrinfo.3.html

```c
#include <limits.h>
#include <stdio.h>
#include <unistd.h>
#include <netdb.h>

int main(void)
{
    int err;

    long host_name_max = sysconf(_SC_HOST_NAME_MAX);
    char hostname[host_name_max + 1];
    
    err = gethostname(hostname, sizeof(hostname));
    if (err != 0) {
        fprintf(stderr, "failed to get hostname: %d", err);
        return 1;
    }

    printf("hostname: %s\n", hostname);

    struct addrinfo hints = {0};
    hints.ai_family = PF_UNSPEC;
    hints.ai_socktype = 0;
    hints.ai_protocol = 0;
    hints.ai_flags = AI_CANONNAME;

    struct addrinfo *info;

    err = getaddrinfo(hostname, NULL, &hints, &info);
    if (err != 0) {
        fprintf(stderr, "failed to get hostname information: %s\n", gai_strerror(err));
        return 1;
    }

    printf("fqdn: %s\n", info->ai_canonname);

    freeaddrinfo(info);

    return 0;
}
```

When running this code, the call to `getaddrinfo` fails with `EAI_NONAME` (_"nodename nor servname provided, or not known"_). This indicates that we call this function incorrectly (either `nodename` or `servname` must be provided but we are clearly doing so).

The run for commit 34441db was on macOS 12.6.9 whereas the run for commit 1a39af0 was on macOS 12.7. There is a chance that some system-level change was introduced that broke us.

PR #103, opened 2025-03-28
Use hostname as primary FQDN source on macOS

bugos:
Fixes: #58 

See explanation in: https://github.com/google/rrg/issues/58#issuecomment-2759571908

### DIFF
--- a/crates/ospect/src/os.rs
+++ b/crates/ospect/src/os.rs
@@ -140,8 +140,6 @@ mod tests {
     }
 
     #[test]
-    // TODO(#58): `ospect::os::fqdn` is broken on macOS.
-    #[cfg_attr(target_os = "macos", ignore)]
     fn fqdn_not_empty() {
         assert!(!fqdn().unwrap().is_empty());
     }
--- a/crates/ospect/src/os/macos.rs
+++ b/crates/ospect/src/os/macos.rs
@@ -31,6 +31,17 @@ pub fn hostname() -> std::io::Result<std::ffi::OsString> {
 }
 
 /// Returns the FQDN of the currently running operating system.
+/// Returns the hostname if it already contains a dot.
 pub fn fqdn() -> std::io::Result<std::ffi::OsString> {
+    let hostname = crate::os::unix::hostname()?;
+
+    // If the hostname contains a dot, it's likely already a FQDN.
+    // It might not be resolvable though so calling
+    // crate::os::unix::fqdn() could fail.
+    use std::os::unix::ffi::OsStrExt;
+    if hostname.as_bytes().contains(&b'.') {
+        return Ok(hostname);
+    }
+
     crate::os::unix::fqdn()
 }


crates/ospect/src/os/unix.rs
```diff
@@ -66,52 +84,44 @@ pub fn fqdn() -> std::io::Result<std::ffi::OsString> {
...
-    if code != 0 {
-        // Ideally, we should use `gai_strerror` to get a human-friendly message
-        // of the error. Unfortunately, it is not clear whether this function is
-        // or is not thread-safe so we just return a generic error.
+    let err = unsafe { libc::getaddrinfo(hostname, std::ptr::null(), &hints, res.as_mut_ptr()) };
+
+    if err != 0 {
+        // SAFETY: gai_strerror returns a pointer to a statically allocated const char *
+        // describing the error.
+        // gai_strerror is thread-safe in MacOS and Linux as long as other threads don't call setenv().
```
panhania, line 99: Do you have a source on that? When I was initially researching this I couldn't find anything about it in [macOS docs][1], nor in [Linux docs][2], nor in [POSIX docs][3].

[1]: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/gai_strerror.3.html
[2]: https://man7.org/linux/man-pages/man3/gai_strerror.3p.html
[3]: https://pubs.opengroup.org/onlinepubs/9799919799/functions/gai_strerror.html

panhania (changes requested):
Hmm, that's a long diff for—from what I understand—is relatively minor change (I think unnecessary formatting changes contribute a lot to that).

But in general, given, that this is only for some incorrect macOS behaviour, cannot we change the implementation of [`fqdn` in the `os/macos`][1] module? On Linux `getaddrinfo` seems to work reliably so using a heuristic of _"dot in the hostname"_ seems like a regression.

And if we modify just the `fqdn` implementation there, there is no need to change any of the existing code nor introduce more `unsafe` stuff there, we can just reuse the high-level functions from `os::unix`. If I understand correctly, refactoring `fqdn` to `resolve_fqdn` was done to avoid repeating a call to `uname` but I think it is not worth the extra complexity, especially if we assume that in majority of modern macOS installations `hostname` will be the correct thing to return without ever doing the second call.

[1]: https://github.com/google/rrg/blob/9677861848ad8fa06696d3f8f3d79dac559f148c/crates/ospect/src/os/macos.rs#L33-L36

crates/ospect/src/os/unix.rs
bugos: I am going to remove the extra error handling for simplicity.

FTR:

I don't have a source but I can see it beeing used in other projects e.g.: https://docs.rs/dns-lookup/latest/src/dns_lookup/addrinfo.rs.html#246

IIUC it takes an error code and returns a statically allocated string, so I guess it is considered trivially thread-safe? With the exception of another thread changing environment variabled concurrently but I understand this is also a problem with `getaddrinfo()`.

panhania (approved):

crates/ospect/src/os/unix.rs
panhania: > IIUC it takes an error code and returns a statically allocated string, so I guess it is considered trivially thread-safe?

The problem is that there is no promise that the string is statically allocated (at least I could not find any confirmation of such). The fact that it does right now should be considered an implementation detail that can change at any time—unless it is documented, I don't think we are allowed to depend on that.

> I don't have a source but I can see it beeing used in other projects

A lot of projects have a very lax approach to writing unsafe code, we try to do better here :D

PR #111, opened 2025-04-16
I add data_truncated in the Result of get_tcp_response.proto since the response data of tcp_response may be truncated in some situation

9-8-7-6:


panhania:
Hm, is this a draft that is intended to be worked on? Introducing this field without any code that would set it seems kind of pointless.

The reason why it is marked as TODO is because the GRR server currently does not care about it and it is not clear whether it will need it. Until there is a need, this seems like adding complexity without any benefit. "YAGNI" and all that...

PR #113, opened 2025-04-18
Add support for pruning when walking directories

panhania:


crates/rrg/src/fs.rs
```diff
@@ -512,4 +575,71 @@ mod tests {
...
+
+    #[test]
+    fn walk_dir_prune_not_dir() {
+        let tempdir = tempfile::tempdir().unwrap();
+        let tempdir = tempdir.path();
+
+        std::fs::create_dir(tempdir.join("abc"))
+            .unwrap();
+        File::create(tempdir.join("abc").join("def")).unwrap();
+        File::create(tempdir.join("abc").join("def")).unwrap();
```
s-westphal, line 604: duplicate line, not sure this is intended?

s-westphal (approved):

PR #117, opened 2025-05-13
Add support for many paths in `get_file_metadata`

panhania:


s-westphal (approved):
Should the filename perhaps indicate more explicitly that metadata of several files can be fetched, e.g. `get_files_metadata` / `get_metadata_for_files`?

panhania:
> Should the filename perhaps indicate more explicitly that metadata of several files can be fetched, e.g. `get_files_metadata` / `get_metadata_for_files`?

Maybe! I filled #119 to look into that one day.

PR #201, opened 2026-03-11
Add filestore support for `scan_memory_yara`

CRefice:
The YARA signature can now either be passed inline along with the action arguments, or uploaded to the filestore ahead of time.

### DIFF
--- a/crates/rrg/src/action/scan_memory_yara.rs
+++ b/crates/rrg/src/action/scan_memory_yara.rs
@@ -11,13 +11,21 @@ use yara_x::blocks::Scanner;
 
 use rrg_proto::scan_memory_yara as proto;
 
+/// Source code of a YARA signature.
+pub enum Signature {
+    /// A YARA signature whose contents are contained directly in the `String` field.
+    Inline(String),
+    /// A YARA signature whose contents must be retrieved from the filestore by its SHA-256.
+    Filestore([u8; 32]),
+}
+
 /// Arguments of the `scan_memory_yara` action.
 pub struct Args {
     /// PIDs of the processes whose memory we are interested in.
     pids: Vec<u32>,
 
     /// YARA signature source to use for scanning.
-    signature: String,
+    signature: Signature,
 
     /// Maximum time spent scanning a single process.
     timeout: Option<Duration>,
@@ -33,6 +41,17 @@ pub struct Args {
     chunk_overlap: u64,
 }
 
+#[derive(Debug)]
+struct MissingSignatureError;
+
+impl std::fmt::Display for MissingSignatureError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "no YARA signature provided")
+    }
+}
+
+impl std::error::Error for MissingSignatureError {}
+
 use crate::request::ParseArgsError;
 impl crate::request::Args for Args {
     type Proto = proto::Args;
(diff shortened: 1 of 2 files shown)


crates/rrg/src/action/scan_memory_yara.rs
```diff
@@ -11,14 +11,22 @@ use yara_x::blocks::Scanner;
 
 use rrg_proto::scan_memory_yara as proto;
 
+/// Source code of a YARA signature.
+pub enum YaraSignature {
+    /// A YARA signature whose contents are contained directly in the `String` field.
+    Inline(String),
+    /// A YARA signature whose contents must be retrieved from the filestore by its SHA-256.
+    FileStore([u8; 32]),
```
panhania: Nit: `Filestore`, not `FileStore` (that's the convention used so far, similarly in the associated comment you written _filestore_ but not _file store_).

proto/rrg/action/scan_memory_yara.proto
```diff
@@ -13,7 +13,12 @@ message Args {
   repeated uint32 pids = 1;
 
   // YARA signature source to use for scanning.
-  string signature = 2;
+  oneof yara_signature {
```
panhania: Nit: I would either:

  * Rename `yara_signature` to `signature_oneof`.
  * Rename `signature` to `signature_inline` (or whatever you decide to go with as the enum option).

Not a fan of using `yara_signature` for oneof and just `signature` for the string at the same time because without looking at the definition, judging purely at the identifiers, you cannot tell a difference between the two—for me they describe the same thing, we just happened to slap `yara_` in front of one of them to make the distinct (but the choice was arbitrary).

Using `signature_oneof`/`signature` or `signature`/`signature_inline` pairs the semantic difference is apparent from the identifiers themselves.

crates/rrg/src/action/scan_memory_yara.rs
```diff
@@ -11,14 +11,22 @@ use yara_x::blocks::Scanner;
 
 use rrg_proto::scan_memory_yara as proto;
 
+/// Source code of a YARA signature.
+pub enum YaraSignature {
+    /// A YARA signature whose contents are contained directly in the `String` field.
+    Inline(String),
```
panhania, line 17: What do you think about _literal_ (no preference on my side, just sharing a thought)?

crates/rrg/src/action/scan_memory_yara.rs
```diff
@@ -40,6 +48,17 @@ pub struct Args {
     chunk_overlap: u64,
 }
 
+#[derive(Debug)]
+struct MissingSignatureError;
```
panhania, line 45: Nit: Given the struct above is named `YaraSignature`, I think this should be consistently named `MissingYaraSignatureError`. Alternatively, `YaraSignature` should be renamed to `Signature` (and I think I am leaning towards this option, given that everything being YARA is kind of implied from the name of the action; it also fits the suggestion about name of the field names below).

proto/rrg/action/scan_memory_yara.proto
CRefice: Fair. I ended up calling the oneof `signature_oneof` and the first variant as `signature_inline`

crates/rrg/src/action/scan_memory_yara.rs
CRefice: I ended up renaming `YaraSignature` to `Signature`.

crates/rrg/src/action/scan_memory_yara.rs
CRefice: It's not a bad idea for the enum variant, but I think `inline` is clearer as a name for the proto field, so I decided to keep it as is for consistency with that.

PR #204, opened 2026-03-18
Add optional annotations to some Result fields for memory actions.

CRefice:
This makes field presence behave as expected.

### DIFF
--- a/proto/rrg/action/dump_process_memory.proto
+++ b/proto/rrg/action/dump_process_memory.proto
@@ -58,37 +58,37 @@ message Result {
   uint32 pid = 1;
 
   // Start offset of the memory region in the process' address space.
-  uint64 region_start = 2;
+  optional uint64 region_start = 2;
   // End offset of the memory region in the process' address space.
-  uint64 region_end = 3;
+  optional uint64 region_end = 3;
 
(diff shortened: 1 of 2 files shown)


panhania:
Could you expand by what you mean by "behave as expected"? In general, I am not fan of `optional` (this is why I stick to `proto3`) and try to design all my protos so that it is not required. So far, we don't have a single `optional` field in the whole RRG codebase, so introducing them here feels weird.

CRefice:
Right, should have expanded. The problem I have is that many of those fields are, quoting the comments, `Set only if `error` is not set.` I think therefore it makes sense to mark those as `optional` both semantically, and so that field presence is set correctly for those fields.

panhania:
> I think therefore it makes sense to mark those as `optional` both semantically (...)

We are using `proto3`, they _are_ optional semantically.

> so that field presence is set correctly for those fields

Do we have a specific use case for using additional field presence of these specific fields? For me using it is a backward compatibility footgun. Otherwise you could argue that you should always just include `optional` but clearly this is not a good idea since it was removed from `proto3` for a reason (and then brought back to accommodate for specific use cases).

CRefice:
Fair enough. I had some use cases in mind, but as I've been writing and rewriting the server-side flow I realized they were unnecessary. I'll close this PR for now.

PR #206, opened 2026-04-08
Add support for `scan_memory_yara` match limit

CRefice:


proto/rrg/action/scan_memory_yara.proto
```diff
@@ -23,6 +23,9 @@ message Args {
   // Maximum time spent scanning a single process.
   google.protobuf.Duration timeout = 3;
 
+  /// Maximum amount of matches returned for each process.
```
panhania: Nit: _"amount"_ is for uncountable, so I think it should be "maximum _number_ of matches".

crates/rrg/src/action/scan_memory_yara.rs
```diff
@@ -275,6 +285,39 @@ fn scan_region<M: MemoryReader>(
...
 
+/// Convert a [yara_x::ScanResults] into a Vec of [Rule]s,
+/// limiting the total amount of matches to `match_limit`.
+fn convert_with_limit(results: yara_x::ScanResults<'_, '_>, match_limit: usize) -> Vec<Rule> {
+    let mut rules = Vec::new();
+    let mut total_matches = 0;
+    for rule in results.matching_rules() {
+        if total_matches >= match_limit {
+            break;
+        }
```
panhania: If you rename `total_matches` to `match_count` you get a nice, common prefix with `match_limit` for nice visual effect.

proto/rrg/action/scan_memory_yara.proto
```diff
@@ -23,6 +23,9 @@ message Args {
...
+  optional uint32 max_matches_per_process = 11;
```
panhania: This should also explain what happens when the limit is crossed. Does the action fail? Does it drop some matches? Which ones?

crates/rrg/src/action/scan_memory_yara.rs
```diff
@@ -275,6 +285,39 @@ fn scan_region<M: MemoryReader>(
...
+        let mut patterns = Vec::new();
+        for pattern in rule.patterns() {
+            if total_matches >= match_limit {
+                break;
+            }
```
panhania: Having two identical `if`s inside nested loops seems redundant. Cannot we use labelled `break` instead?

crates/rrg/src/action/scan_memory_yara.rs
```diff
@@ -275,6 +285,39 @@ fn scan_region<M: MemoryReader>(
     Ok(())
 }
 
+/// Convert a [yara_x::ScanResults] into a Vec of [Rule]s,
+/// limiting the total amount of matches to `match_limit`.
+fn convert_with_limit(results: yara_x::ScanResults<'_, '_>, match_limit: usize) -> Vec<Rule> {
```
panhania: Not a fan of this name as it does not really convey much meaning. Instead of `convert`, prefer something like `x_to_y`. Here it is going to be quite mouthful but I think that is better than having unhelpful identifiers at all.

panhania (changes requested):

proto/rrg/action/scan_memory_yara.proto
panhania: There are some examples of this in other files as well.

proto/rrg/action/scan_memory_yara.proto
panhania: Also, the way it is implemented, it is going to omit reporting not only matches but also rules that matched. Is this expected?

CRefice:
So I thought about your comments and figured that:
1. The parameter is only really meant to be used for testing (e.g. testing out a new signature) to keep runtime down and see if it works at all
2. Since yara-x only allows setting a maximum amount of matches per pattern (instead of per-process), limiting the amount of results after scanning is complete doesn't really keep runtime down, if anything it increases it.

Therefore I decided to change course and change the parameter to `max_matches_per_pattern` and just propagate it to yara and that's it. This simplifies the logic a lot.

PR #207, opened 2026-04-17
Use non-blocking reads in `get_file_metadata`

panhania:
No tests as apparently there is no reliable way to do it—we can create only non-blocking regular files and such are not considered for reading anyway. `/proc/kmsg` that has the issue requires root privileges to be read so that also is a no-go. Given that the code is covered anyway (it always executes), I think that is tolerable omission.

### DIFF
--- a/crates/rrg/src/action/get_file_metadata.rs
+++ b/crates/rrg/src/action/get_file_metadata.rs
@@ -157,7 +157,19 @@ where
                     buf.capacity(),
                 };
 
-                let mut file = match std::fs::File::open(&entry.path) {
+                let mut file_open_opts = std::fs::OpenOptions::new();
+                file_open_opts.read(true);
+
+                // Some regular files can be blocking (e.g. `/proc/kmsg` on
+                // Linux) so we do only non-blocking reads to avoid process
+                // hangups.
+                #[cfg(target_family = "unix")]
+                {
+                    use std::os::unix::fs::OpenOptionsExt as _;
+                    file_open_opts.custom_flags(libc::O_NONBLOCK);
+                }
+
+                let mut file = match file_open_opts.open(&entry.path) {
                     Ok(file) => file,
                     Err(error) => {
                         log::error! {
@@ -307,7 +319,18 @@ fn digest(path: &Path, metadata: &std::fs::Metadata, args: &Args) -> Digest {
         return Digest::default();
     }
 
-    let mut file = match std::fs::File::open(path) {
+    let mut file_open_opts = std::fs::OpenOptions::new();
+    file_open_opts.read(true);
+
+    // Some regular files can be blocking (e.g. `/proc/kmsg` on Linux) so we do
+    // only non-blocking reads to avoid process hangups.
+    #[cfg(target_family = "unix")]
+    {
+        use std::os::unix::fs::OpenOptionsExt as _;
+        file_open_opts.custom_flags(libc::O_NONBLOCK);
+    }
+
+    let mut file = match file_open_opts.open(path) {
         Ok(file) => std::io::BufReader::new(file),
         Err(error) => {
             log::error!("failed to open '{}' for digest: {error}", path.display());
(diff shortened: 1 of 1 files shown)


crates/rrg/src/action/get_file_metadata.rs
```diff
@@ -157,7 +157,19 @@ where
                     buf.capacity(),
                 };
 
-                let mut file = match std::fs::File::open(&entry.path) {
+                let mut file_open_opts = std::fs::OpenOptions::new();
+                file_open_opts.read(true);
+
+                // Some regular files can be blocking (e.g. `/proc/kmsg` on
```
CRefice, line 163: Reading the [man page for O_NONBLOCK](https://linux.die.net/man/3/open#:~:text=O_NONBLOCK,O_NONBLOCK%20is%20unspecified.), it seems like using `O_NONBLOCK` has an unspecified effect for regular files. I suppose you meant special files?
It's also not clear to me what happens after the file is opened. reading could still block with this flag, from what I gather from the man page. What is the intended usecase of this?

crates/rrg/src/action/get_file_metadata.rs
```diff
@@ -307,7 +319,18 @@ fn digest(path: &Path, metadata: &std::fs::Metadata, args: &Args) -> Digest {
...
+
+    // Some regular files can be blocking (e.g. `/proc/kmsg` on Linux) so we do
+    // only non-blocking reads to avoid process hangups.
+    #[cfg(target_family = "unix")]
+    {
+        use std::os::unix::fs::OpenOptionsExt as _;
+        file_open_opts.custom_flags(libc::O_NONBLOCK);
+    }
+
+    let mut file = match file_open_opts.open(path) {
```
CRefice, line 335: I know you prefer inlining logic, but since this block of code is exactly identical to the one above, down to the comment and everything, maybe consider outlining this into a function. Up to you.

crates/rrg/src/action/get_file_metadata.rs
panhania: > I suppose you meant special files?

Nope, I meant regular. Non-regular files are exempt from content matching and thus are never opened.

> It's also not clear to me what happens after the file is opened. reading could still block with this flag, from what I gather from the man page. What is the intended usecase of this?

Because there are apparently files such as `/proc/kmsg` that are "regular" (at least that is what the systems says) and blocking, thus hanging the process when you attempt to read from it. `O_NONBLOCK` seems to work for it where we get an error instead of a hangup.

crates/rrg/src/action/get_file_metadata.rs
panhania: I find it hard to come up with abstraction for this case:

  * I cannot name it `open` because that immediately raises a question: why is there some `open` used instead of the standard one thus forcing the reader to jump to the definition.
  * I cannot name it `open_nonblocking` because the nonblocking behaviour is UNIX-specific.
  * I maybe could name if `open_nonblocking_on_unix` but then the name essentially repeats the implementation but in a poor way as it is not obvious what this does on Windows (does it fail? does it just use normal open?).
  * I could expand the name even further but we just go more toward repeating the implementation inside the name.

And thus, if I cannot sensibly name something, I prefer not name it at all.

With that being said, there is a more jarring issue here: we open the file in two places. I plan to refactor the code in the future that this is not the case and thus the duplication should also go away with it.

crates/rrg/src/action/get_file_metadata.rs
CRefice: Oh okay, so the intended behavior is that it returns an error if it tries to block. I'd specify that in this comment.

panhania: > I'd specify that in this comment.

Done.

crates/rrg/src/action/get_file_metadata.rs
panhania: > I plan to refactor the code in the future that this is not the case and thus the duplication should also go away with it.

I filed #209 not to forget about this.

CRefice (approved):

PR #211, opened 2026-05-13
Add inline transfer mode in `get_file_contents`

panhania:


CRefice (approved):
Q: I think this is something that might come up for other flows as well. Would it make sense to either move this logic to the transport layer, or at least into a library?

panhania:
> Q: I think this is something that might come up for other flows as well. Would it make sense to either move this logic to the transport layer, or at least into a library?

Maybe! Hard to tell before we have multiple use cases where we clearly see some pattern.

So far the only uses of the blob sink seems to be: `get_file_contents`, `scan_memory_yara` and `dump_process_memory`. It's difficult for me to see why `scan_memory_yara` and `dump_process_memory` would want that (given that matches can be arbitrarily big)—do you have some other flow in mind or do you see some fit for these?

CRefice:
Yeah, that's what I was referring to. I could foresee the pattern of "send back a blob of data, maybe inline if it's small enough or I ask for it explictily" to be useful for those other two as well. But anyway, I was just thinking out loud, we can think about that if and when we get there.

panhania:
> I could foresee the pattern of "send back a blob of data, maybe inline if it's small enough or I ask for it explictily" to be useful for those other two as well.

Yeah, I guess I could see it as a potential optimization but given that would complicate the logic (you have to support both sink and non-sink) I would not touch it until we really have nothing to work on :D

The main motivator here is to be used for flows that collect files that we know are small (e.g. procfs stuff or `/etc/lsb-release` information) and so I can use only the inline mode, letting me shave a lot of logic related to blob awaits and stuff.

PR #214, opened 2026-06-12
Implement process metadata collection for macOS

panhania:
This is just the part that goes into the `ospect` crate. Implementation of the action that uses it will follow shortly.

### DIFF
--- a/crates/ospect/src/proc/macos.rs
+++ b/crates/ospect/src/proc/macos.rs
@@ -334,7 +334,7 @@ mod tests {
     use super::*;
 
     #[test]
-    fn all_self_exists() {
+    fn all_self() {
         let metadata = all()
             .unwrap().filter_map(Result::ok)
             .find(|metadata| metadata.id() == std::process::id())
@@ -354,7 +354,7 @@ mod tests {
     }
 
     #[test]
-    fn all_launchd_exists() {
+    fn all_launchd() {
         let metadata = all()
             .unwrap().filter_map(Result::ok)
             .find(|metadata| metadata.id() == 1)
--- a/crates/ospect/src/proc/macos.rs
+++ b/crates/ospect/src/proc/macos.rs
@@ -5,7 +5,18 @@
 
 /// Returns an iterator yielding identifiers of all processes on the system.
 pub fn ids() -> std::io::Result<impl Iterator<Item = std::io::Result<u32>>> {
-    Ok(all()?.map(|metadata| Ok(metadata?.id())))
+    Ok(sysctl_kern_proc()?.into_iter().map(|proc| {
+        u32::try_from(proc.kp_proc.p_pid)
+            .map_err(|error| std::io::Error::new(
+                std::io::ErrorKind::InvalidData,
+                error,
+            ))
+    }))
+}
+
+/// Returns an iterator yielding metadata for all processes on the system.
+pub fn all() -> std::io::Result<impl Iterator<Item = std::io::Result<Metadata>>> {
+    Ok(sysctl_kern_proc()?.into_iter().map(|raw| Ok(Metadata { raw })))
 }
 
 /// Metadata about the process (specific to macOS).
@@ -81,11 +92,6 @@ impl Metadata {
             argc_left: argc,
         })
     }
-}
-
-/// Returns an iterator yielding metadata for all processes on the system.
-pub fn all() -> std::io::Result<impl Iterator<Item = std::io::Result<Metadata>>> {
-    Ok(sysctl_kern_proc()?.into_iter().map(|raw| Ok(Metadata { raw })))
 }
 
 /// Iterator over the arguments of the process.
--- a/crates/ospect/src/proc/macos.rs
+++ b/crates/ospect/src/proc/macos.rs
@@ -48,6 +48,102 @@ impl Metadata {
 
         std::ffi::OsStr::from_bytes(name_bytes).to_os_string()
     }
+
+    pub fn args(&self) -> std::io::Result<Args> {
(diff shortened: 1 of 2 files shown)


crates/ospect/src/proc/macos.rs
```diff
@@ -5,116 +5,399 @@
...
+
+        assert_eq!(metadata.parent_id(), 0);
+        assert_eq!(metadata.name(), "launchd");
+    }
+
+    #[test]
+    fn all_subprocess() {
+        // TODO(rust-lang/rust#144426): Simplify once `drop_guard` is stable.
+        struct ChildDropGuard(std::process::Child);
+        impl Drop for ChildDropGuard {
```
CRefice, line 361: Missed opportunity for some kind of terrible joke about killing children :) /s

crates/ospect/src/proc/macos.rs
```diff
@@ -5,116 +5,399 @@
...
-        };
-        if code != 0 {
-            return Err(std::io::Error::last_os_error());
-        }
+    /// Returns PID of the process this metadata corresponds to.
+    pub fn id(&self) -> u32 {
+        u32::try_from(self.proc.kp_proc.p_pid)
+            // PID for a live process should never be negative and sign is used
+            // only for special return values or arguments.
+            .unwrap_or(0)
```
CRefice, line 44: I don't fully understand the context of the comment (return values or arguments to what?) but if it's never supposed to happen, wouldn't a straight `unwrap` help catch a mistake sooner? Or at least a log instead of just returning `0`? Same for `parent_pid` below.

crates/ospect/src/proc/macos.rs
```diff
@@ -5,116 +5,399 @@
...
-        // The rest is as with the `sysctl` call above (not the comment about
-        // the length parameters).
-        let code = unsafe {
-            libc::sysctl(
-                mib.as_mut_ptr(), mib.len() as libc::c_uint,
-                buf.as_mut_ptr().cast::<libc::c_void>(), &mut buf_size,
-                std::ptr::null_mut(), 0,
-            )
+        let name_bytes = self.proc.kp_proc.p_comm.map(|byte| byte as u8);
+        // Name is null-terminated so we need to take it only until the null
```
CRefice, line 60: I think this might be reimplementing logic from [`CStr::from_bytes_until_nul`](https://doc.rust-lang.org/std/ffi/struct.CStr.html#method.from_bytes_until_nul)

crates/ospect/src/proc/macos.rs
```diff
@@ -5,116 +5,399 @@
...
-        // the buffer above.
-        unsafe {
-            buf.set_len(buf_len);
-        }
+        // We start at index `4` because of the 4-byte argc value placed at the
+        // beginning of the buffer.
+        let exe_start = 4;
+        // Executable name is null-terminated so we find the position of the
+        // first null byte since the beginning.
+        let exe_len = proc_args[exe_start..].iter().position(|byte| *byte == 0)
```
CRefice: See above

crates/ospect/src/proc/macos.rs
```diff
@@ -5,116 +5,399 @@
...
+    // SAFETY: If the call to `sysctl` succeeded, we can assume that the
+    // `buf_size` no is filled with the expected size of the buffer.
+    let mut buf_size = unsafe {
+        buf_size.assume_init()
+    } as usize;
+
+    let mut buf_len = buf_size / KINFO_PROC_SIZE;
+    if buf_size % KINFO_PROC_SIZE != 0 {
+        buf_len += 1;
+    }
```
CRefice: Nit: if our toolchain targets rust >= 1.73 we can use [`usize::div_ceil`](https://doc.rust-lang.org/std/primitive.usize.html#method.div_ceil) for a clearer intent

CRefice (approved):

crates/ospect/src/proc/macos.rs
panhania: Oh, so obvious when I look at it now...

crates/ospect/src/proc/macos.rs
panhania: >  (return values or arguments to what?)

I clarified the comment.

> if it's never supposed to happen, wouldn't a straight `unwrap` help catch a mistake sooner?

I am not entirely sure if this assumption is guaranteed in any standard so that I could quote it why it cannot panic. Nevertheless, it is assumption not in the code we own so I think it is better to play it safe here.

> Or at least a log instead of just returning `0`?

We cannot, the `ospect` crate shouldn't do any side-effects (the intention is that it is completely independent from RRG).

crates/ospect/src/proc/macos.rs
panhania: The funny thing is that I even started with it, then I noticed that `p_comm` is `&[libc::char]` (not `&[u8]`) and didn't work, so I dropped it and worked with `.position` as comparison to 0 works just fine. But then I hit some other type shenanigans and eventually added `.map((...) as u8)` above but never went back to `CStr::from_bytes_*`.

Looks much more obvious now, thanks.

crates/ospect/src/proc/macos.rs
panhania: Also done!

crates/ospect/src/proc/macos.rs
panhania: I did not know about `usize::div_ceil`, I agree it is clearer!

We can be pretty aggressive about raising MSRV (the internal toolchain even tracks `beta`, I think).

PR #218, opened 2026-06-18
Implement abort reporting

panhania:
This implements a mechanism similar to the one found in the old agent where we have a [`Client.transaction_log_file`](https://github.com/google/grr/blob/9f8f7974b3ba063a7e60526d4881db0c88746aa1/grr/core/grr_response_core/config/client.py#L154-L158) that persists [currently processes messages](https://github.com/google/grr/blob/9f8f7974b3ba063a7e60526d4881db0c88746aa1/grr/client/grr_response_client/client_utils_osx_linux.py#L97-L117). The file is cleared once [we are done processing](https://github.com/google/grr/blob/9f8f7974b3ba063a7e60526d4881db0c88746aa1/grr/client/grr_response_client/comms.py#L215-L216). If the agent dies mid-processing, the file is not removed and can be used to send information about the aborted operation back to the GRR server which [marks the requesting flow as failed](https://github.com/google/grr/blob/9f8f7974b3ba063a7e60526d4881db0c88746aa1/grr/server/grr_response_server/flows/general/administrative.py#L120-L127).

Currently, if RRG dies, the server never gets the response back to the action request and "hangs" forever which is not a great user experience.

In RRG we use slightly different nomenclature:

  * _Request file_ rather than _transaction log_, as I think it better captures the structure and the purpose of the file.
  * _Abort_ rather than _crash_ as the agent can die to various reasons, including scheduled Fleetspeak or system restarts.

### DIFF
--- a/crates/rrg/src/abort.rs
+++ b/crates/rrg/src/abort.rs
@@ -56,14 +56,16 @@ impl RequestFileCreated {
     ///
     /// This should be only used if the request processing has been finished.
     pub fn remove(self) -> std::io::Result<()> {
-        // TODO(@panhania): Still attempt to delete the file even if unlock
-        // fails.
-        self.file.unlock()?;
+        // We still want to attempt to remove the file, even if unlocking it
+        // (for whatever reason) fails, so we don't immediately propagate the
+        // error but still report it at the end.
+        let result_unlock = self.file.unlock();
+
         drop(self.file);
 
         std::fs::remove_file(&self.path)?;
 
-        Ok(())
+        result_unlock
     }
 }
 
--- a/crates/rrg/src/abort.rs
+++ b/crates/rrg/src/abort.rs
@@ -1,0 +1,253 @@
+// Copyright 2026 Google LLC
+//
+// Use of this source code is governed by an MIT-style license that can be found
+// in the LICENSE file or at https://opensource.org/licenses/MIT.
+
+//! Utilities for abort reporting.
+
+/// Handle to a file with persisted request for abort recovery.
+///
+/// This is created with the [`create_request_file`] function and should be used
+/// by request processing logic.
+#[derive(Debug)]
+pub struct RequestFileCreated {
+    /// Path to the request file.
+    path: std::path::PathBuf,
+    /// Created request file (used to hold the lock).
+    file: std::fs::File,
+}
+
+/// Persists given request information at the specified path.
+///
+/// Once request processing is finished, [`RequestFileCreated::remove`] should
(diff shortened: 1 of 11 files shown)


CRefice:
Can you share some details about the purpose and high-level logic of this feature in this PR? Just so I have something to compare to when looking at the code

panhania:
I updated the PR description.

proto/rrg/abort.proto
```diff
@@ -0,0 +1,40 @@
...
+// in the LICENSE file or at https://opensource.org/licenses/MIT.
+syntax = "proto3";
+
+package rrg.abort;
+
+import "google/protobuf/timestamp.proto";
+
+// Subset of the request information used for abort reporting.
+//
+// This is information is persisted in a file before request processing is
```
CRefice:
```suggestion
// This information is persisted in a file before request processing is
```

crates/rrg/src/abort.rs
```diff
@@ -0,0 +1,255 @@
...
+        // not be removed.
+        let request_file = open_request_file(&request_file_path)
+            .unwrap();
+
+        assert_eq!(request_file.flow_id, request_id.flow_id());
+        assert_eq!(request_file.request_id, request_id.request_id());
+        // We cannot use more concrete timestamps (e.g. before and after the
+        // computation thread was started) because the filesystem clock and the
+        // system time might not be in a perfect sync.
+        assert!(request_file.request_time >= std::time::UNIX_EPOCH);
```
CRefice: I get the reasoning behind the comment, but this seems like an excessively wide test. After all, a default `request_time` be equal to UNIX_EPOCH (i.e. zero), right?
At least making it strictly greater than might be a more useful test.

Also now that I think about it, can't the issues with clock drift be avoided by just serializing system time in the `Request` as well?

CRefice (approved):

crates/rrg/src/abort.rs
panhania: > At least making it strictly greater than might be a more useful test.

Uh, yeah, the strict inequility was the intended behavior. The weak version stuck here because the code was adapted from the previous timestamp-based tests. Adjusted!

> Also now that I think about it, can't the issues with clock drift be avoided by just serializing system time in the `Request` as well?

During implementation I wend back and forth on doing it or not. I think the desire for minimalism and not duplicating data when needed prevailed.

But even if we stuck with the other approach, serializing system time won't solve the issue completely as `std::time::SystemTime` is not guaranteed to be monotonic (only `std::time::Instant` is).

PR #222, opened 2026-06-26
Report timeouts in `execute_signed_command`

CRefice:
Add a new `timed_out` field to `Result` for the `execute_signed_command` action, set to true if the subcommand exceeded the specified timeout. This is particularly useful on Windows where we didn't report any information on _how_ the process exited (like we do with signals on Unix), so a timeout would just result in an exit code of 1 and an empty stdout/stderr.

### DIFF
--- a/proto/rrg/action/execute_signed_command.proto
+++ b/proto/rrg/action/execute_signed_command.proto
@@ -155,5 +155,13 @@ message Result {
 
   // Set if value of `stderr` had to be truncated.
   bool stderr_truncated = 6;
+
+  // Set if the command timed out before finishing.
+  //
+  // This does not necessarily mean the command failed,
+  // as the process could have exited normally in the time
+  // between when the timeout was detected and when the process
+  // was killed.
+  bool timeout_reached = 7;
 }
 
--- a/crates/rrg/src/action/execute_signed_command.rs
+++ b/crates/rrg/src/action/execute_signed_command.rs
@@ -40,6 +40,8 @@ pub struct Item {
     stderr: Vec<u8>,
     /// Whether stderr is truncated.
     truncated_stderr: bool,
+    /// Whether the command reached the specified timeout before exiting.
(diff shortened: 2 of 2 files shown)


proto/rrg/action/execute_signed_command.proto
```diff
@@ -155,5 +155,9 @@ message Result {
 
   // Set if value of `stderr` had to be truncated.
   bool stderr_truncated = 6;
+
+  // Set if the command timed out before finishing.
+  // This does not necessarily mean the command failed.
+  bool timed_out = 7;
```
panhania: Could we rename it to `timeout_reached` (or `timeout_exceeded` as you use in the log message, although "reaching" a timeout feels more grammatically sound, I think _time_ can be exceeded)?

It is a bit lengthier by I feel like it has a nice property that in the arguments we have `timeout` and then we have this `timeout_reached` that corresponds to it and differs only in a suffix. Whereas `timed_out` has completely different structure to it.

proto/rrg/action/execute_signed_command.proto
```diff
@@ -155,5 +155,9 @@ message Result {
 
   // Set if value of `stderr` had to be truncated.
   bool stderr_truncated = 6;
+
+  // Set if the command timed out before finishing.
+  // This does not necessarily mean the command failed.
```
panhania: Nit: I tend to have a single-sentence header on the first line and then (optionally) more supplementary documentation separated from the header by a blank link).

proto/rrg/action/execute_signed_command.proto
CRefice: Done, I uniformed the changes to use "reaching" as a wording consistently.

proto/rrg/action/execute_signed_command.proto
CRefice: Done.

panhania (approved):

PR #224, opened 2026-07-06
Use inodes for listing connections on Linux

CRefice:
The current implementation of ospect's network connection listing functionality on Linux tries to read a process' open connections from procfs, for instance from `/proc/<pid>/net/tcp`. However, `/proc/<pid>/net/tcp` is effectively the same as `/proc/net/tcp`, meaning it lists _all_ open connections on the system, whether they were opened by the process in question or not. This results in incorrect process attribution, and even worse, in `all_connections()` returning multiple copies of the results, one per active process on the system!

To fix this, this PR introduces a similar strategy to that used by the python `psutil` package which was used by the old GRR agent: parsing the connection's inode from `/proc/net/`, and matching it with the process' currently open file descriptors (from `/proc/<pid>/fd`), ignoring the result if there's no match.

This can still result in some duplicate results being returned, since for example, a process can open a socket and then fork, in which case the connection will be considered as "owned" by both parent and child. But in the general case, this will no longer result in hundreds of thousands of duplicate results.

Overall, the public interface of `net.rs` is not well suited to Linux, where there is no 1:1 ownership between a process and a network connection. Having a standalone function to list connections owned by a given process means we must scan `/proc/net/` once for each process, which is not ideal for performance.
Changing that would require more extensive changes to the implementation for other platforms as well, so we can save it for a future improvement.

### DIFF
--- a/crates/ospect/src/net.rs
+++ b/crates/ospect/src/net.rs
@@ -666,7 +666,6 @@ mod tests {
 
         let mut conns = tcp_v4_connections(std::process::id())
             .unwrap()
-            .map(|x| dbg!(x))
             .filter_map(Result::ok);
 
         let server_conn = conns
@@ -752,22 +751,25 @@ mod tests {
         let udp_socket_addr = udp_socket.local_addr()
             .unwrap();
 
-        let mut conns = all_connections()
+        let conns = all_connections()
             .unwrap()
-            .filter_map(Result::ok);
+            .filter_map(Result::ok)
+            .collect::<Vec<_>>();
 
-        assert! {
-            conns.find(|conn| {
+        // Check that the sockets are found and that there are no duplicates
+
+        assert_eq!(
+            conns.iter().filter(|conn| {
                 conn.pid() == std::process::id() &&
                 conn.local_addr() == tcp_server_addr
-            }).is_some()
-        }
+            }).count(), 1
+        );
 
-        assert! {
-            conns.find(|conn| {
+        assert_eq!(
+            conns.iter().filter(|conn| {
                 conn.pid() == std::process::id() &&
                 conn.local_addr() == udp_socket_addr
-            }).is_some()
-        }
+            }).count(), 1
+        )
     }
 }
--- a/crates/ospect/src/net.rs
+++ b/crates/ospect/src/net.rs
@@ -164,6 +164,9 @@ struct TcpConnectionInner<A> {
     state: TcpState,
     /// An identifier of the process that owns the connection.
     pid: u32,
+    // inode identifier of the open socket file descriptor.
+    #[cfg(target_os = "linux")]
+    inode: u64,
 }
 
 /// Information about a TCP IPv4 connection.
@@ -315,6 +318,9 @@ pub struct UdpConnectionInner<A> {
     local_addr: A,
     /// An identifier of the process that owns the connection.
     pid: u32,
+    // inode identifier of the open socket file descriptor.
+    #[cfg(target_os = "linux")]
+    inode: u64,
 }
 
 /// Information about a UDP IPv4 connection.
@@ -660,6 +666,7 @@ mod tests {
 
         let mut conns = tcp_v4_connections(std::process::id())
             .unwrap()
+            .map(|x| dbg!(x))
             .filter_map(Result::ok);
 
         let server_conn = conns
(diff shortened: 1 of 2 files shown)


s-westphal (approved):

crates/ospect/src/net.rs
```diff
@@ -660,6 +666,7 @@ mod tests {
 
         let mut conns = tcp_v4_connections(std::process::id())
             .unwrap()
+            .map(|x| dbg!(x))
```
panhania: `dbg!`

crates/ospect/src/net/linux/conn.rs
```diff
@@ -3,42 +3,94 @@
 // Use of this source code is governed by an MIT-style license that can be found
 // in the LICENSE file or at https://opensource.org/licenses/MIT.
 
+use std::collections::HashSet;
+
 use crate::net::*;
 
+/// Returns a set of inodes of the given process' currently open sockets, parsed from procfs.
+/// The return value be used to cross-check whether a given connection "belongs" to the process.
+fn get_open_socket_inodes(pid: u32) -> std::io::Result<HashSet<u64>> {
```
panhania: Can we move this below? I prefer to have public or most prominent methods at the top of the file with utilities towards the bottom.

crates/ospect/src/net/linux/conn.rs
```diff
@@ -3,42 +3,94 @@
...
+            .strip_prefix("socket:[")
+            .and_then(|str| str.strip_suffix(']'));
+        // Be lenient and treat a failure to parse an integer as the inode being absent.
+        // In practice, this should never happen.
+        inode_num.and_then(|s| s.parse().ok())
+    }
+
+    let mut results = HashSet::new();
+    for entry in std::fs::read_dir(format!("/proc/{pid}/fd"))?.filter_map(Result::ok) {
+        // readlink can fail
```
panhania: Nit: Missing dot at the end of the comment. I would also backtick `read_link` (and include underscore) as it refers to the Rust's standard library function, not to the Bash utility.

crates/ospect/src/net/linux/conn.rs
```diff
@@ -3,42 +3,94 @@
...
+        let Ok(resolved_path) = dbg!(std::fs::read_link(entry.path())) else {
```
panhania: `dbg!`

crates/ospect/src/net/linux/conn.rs
```diff
@@ -3,42 +3,94 @@
...
+            continue;
+        };
+        let Some(path_str) = resolved_path.to_str() else {
+            // Fds can point to arbitrary disk paths which may not be UTF-8,
+            // but we don't care about those.
+            continue;
+        };
+        if let Some(inode) = parse_socket_inode(path_str) {
+            results.insert(inode);
+        }
```
panhania: Nit: For better flow with other checks, I would also use `else { continue }` and have a `result.insert(inode)` at the loop level. The sudden change of style is a bit confusing on the first sight.

crates/ospect/src/net/linux/conn.rs
```diff
@@ -3,42 +3,94 @@
...
+/// Returns a set of inodes of the given process' currently open sockets, parsed from procfs.
+/// The return value be used to cross-check whether a given connection "belongs" to the process.
+fn get_open_socket_inodes(pid: u32) -> std::io::Result<HashSet<u64>> {
+    fn parse_socket_inode(path_str: &str) -> Option<u64> {
+        let inode_num = path_str
+            .strip_prefix("socket:[")
+            .and_then(|str| str.strip_suffix(']'));
+        // Be lenient and treat a failure to parse an integer as the inode being absent.
+        // In practice, this should never happen.
+        inode_num.and_then(|s| s.parse().ok())
```
panhania: Couple nits: 
* I find `inode_num` being an `Option` slightly burried (one can deduce this only from `.and_then`). I would  either remove the variable altogether and chain `.and_then` or I would use `?` to keep `inode_num` a concrete value.
* In one closure you use `str`, in the other `s`, I would keep it consistent (with personal preference towards `str`).
* `.parse().ok()` is a bit unobvious what you are parsing into (one has to jump to the function signature). I would turbofish it to `.parse::<u64>().ok()` to make it a bit more explicit.

panhania (changes requested):

panhania (approved):

PR #233, opened 2026-08-28
Fix fuzz harness aborts and timeouts

pmateatg:


fuzz/fuzz_targets/fuzz_action_execute_signed_command.rs
```diff
@@ -44,6 +44,14 @@ struct FuzzInput {
 }
 
 fuzz_target!(|input: FuzzInput| {
+    if input.command.args.len() > 16
+        || input.command.env.len() > 16
+        || input.unsigned_args.len() > 16
+        || input.unsigned_env.len() > 16
```
panhania: It would be nice if libfuzzer supported something similar to quickcheck's discarding feature ([example usage](https://github.com/google/rrg/blob/1331b2ef7e14532195ceb4e7940a6d56e0c4400e/crates/rrg/src/io.rs#L318-L322)). This allows to guide the sample picker and skips huge space of uninteresting inputs.

fuzz/fuzz_targets/fuzz_action_get_file_contents.rs
```diff
@@ -20,6 +20,10 @@ struct FuzzInput {
 }
 
 fuzz_target!(|input: FuzzInput| {
+    if input.file_content.len() > 64 * 1024 {
+        return;
+    }
```
panhania: Any insights on how these limits were picked? I see some targets use 64 KiB, some use 32 KiB.

fuzz/fuzz_targets/fuzz_action_execute_signed_command.rs
```diff
@@ -85,7 +93,9 @@ fuzz_target!(|input: FuzzInput| {
     let mut proto = RequestProto::new();
     proto.set_request_id(12345);
     proto.set_action(rrg_proto::rrg::Action::EXECUTE_SIGNED_COMMAND);
-    proto.set_args(protobuf::well_known_types::any::Any::pack(&args).unwrap());
+    if let Ok(args_any) = protobuf::well_known_types::any::Any::pack(&args) {
+        proto.set_args(args_any);
+    }
```
panhania: I wonder why this is needed. I though every valid proto message should be "packable".

fuzz/fuzz_targets/fuzz_action_get_tcp_response.rs
```diff
@@ -41,7 +45,25 @@ fuzz_target!(|input: FuzzInput| {
...
+                Ok((s, _)) => break Some(s),
+                Err(ref e) if e.kind() == std::io::ErrorKind::WouldBlock => {
+                    if start.elapsed() > std::time::Duration::from_millis(50) {
+                        break None;
+                    }
+                    thread::sleep(std::time::Duration::from_millis(1));
+                }
+                Err(_) => break None,
+            }
+        };
```
panhania: Could you include a comment what situations we guard ourselves against here? It feels like the code to establish the connection is pretty deterministic so it should eventually always accept (unless the main thread hangs or panics in which case it is also not clear what this buys us here).

panhania (changes requested):

fuzz/fuzz_targets/fuzz_action_execute_signed_command.rs
pmateatg: The fuzzer is coverage guided, it's not bruteforcing all the possible inputs, but it can be nudged in the right way with a good corpus and/or a dictionary of keywords.

On the other hand, I created BoundedVec in fuzz_utils to make the limits look more intentional and less look like a band aid fix. When the fuzzer sees it's not finding new edges with longer inputs it'll stop trying. For example:

```
struct FuzzCommand {
    path: String,
    args: BoundedVec<FuzzCommandArg, MAX_FUZZ_VEC_LEN>,
    env: BoundedVec<(String, String), MAX_FUZZ_VEC_LEN>
}
```

fuzz/fuzz_targets/fuzz_action_get_file_contents.rs
pmateatg: With the BoundedVec change I also moved the magic values to fuzz_utils as constants, so they are the same across the board.

They are essentially empirical values. RRG only imposes limits on the client resources usage by the Session, which is bypassed by the fuzzer, we need to impose some limits. The number can't be too large because it'll translate to longer runtime, and can't be too low that it hinders the fuzzer's efforts.

fuzz/fuzz_targets/fuzz_action_execute_signed_command.rs
pmateatg: Reverted these changes, got mixed up with Proto2 and Proto3, in proto3 everything is "packable".

fuzz/fuzz_targets/fuzz_action_get_tcp_response.rs
pmateatg: The issue is that the connect_timeout is controlled by the fuzzer, and the connection is not established  when the fuzzer sets it to 0 and leaves the server hanging.

After understanding this better I just chose to ensure the connection timeout is always positive, so we don't need to defend against unrealistic fuzzer generated scenarios.

pmateatg:
Please take a look again, it's largely the same just with cleanups and more comments.

panhania (approved):
Looks good, I will merge after some other pending PRs are in.

PR #243, opened 2026-08-28
Close enumerator on happy path when querying WMI

grrrrrrrrr:
In `wmi::QueryRows::next`, property iteration begins with
`IWbemClassObject::BeginEnumeration` and iterates using `Next`. When
`Next` returned `WBEM_S_NO_MORE_DATA`, the loop terminated without
calling `EndEnumeration`.

Microsoft WMI COM documentation specifies that `EndEnumeration` must be
called to release internal enumeration resources before starting a new
enumeration or releasing the object.

This change adds the missing `EndEnumeration` call upon loop completion.

### DIFF
--- a/crates/wmi/src/windows.rs
+++ b/crates/wmi/src/windows.rs
@@ -286,6 +286,14 @@ impl<'com> Iterator for QueryRows<'com> {
             row.insert(name, value);
         }
 
+        // SAFETY: EndEnumeration must be called when enumeration is completed.
+        let status = unsafe {
+            (object.vtable().EndEnumeration)(object.as_raw_mut())
+        };
+        if status != windows_sys::Win32::Foundation::S_OK {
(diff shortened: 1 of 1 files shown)


panhania (approved):

panhania:
> Microsoft WMI COM documentation specifies that `EndEnumeration` must be
> called to release internal enumeration resources before starting a new
> enumeration or releasing the object.

That is actually not the case, per [`EndEnumeration` docs](https://learn.microsoft.com/en-us/windows/win32/api/wbemcli/nf-wbemcli-iwbemclassobject-endenumeration):

> This call is not required, but it is recommended to developers because it releases resources associated with the enumeration. However, the resources are deallocated automatically when the next enumeration is started or the object is released.

Still, makes no harm to release it explicitly.

PR #248, opened 2026-08-28
Retry process spawn on `ExecutableFileBusy` in `execute_signed_command`

grrrrrrrrr:
When executing scripts or binaries freshly written to disk (such as files assembled from the filestore), calling `std::process::Command::spawn` on Unix can fail with `ETXTBSY` (`std::io::ErrorKind::ExecutableFileBusy`, "Text file busy"). This occurs when parallel threads fork child processes concurrently while a writable file descriptor is open, causing the child process to temporarily inherit the open file descriptor until `execve` runs or the `CLOEXEC` descriptor is closed.

This change adds a brief retry loop on `ErrorKind::ExecutableFileBusy` (up to 5 attempts with 10 ms backoff) when spawning commands in `execute_signed_command`, eliminating flaky failures during concurrent execution and multi-threaded test runs.

### DIFF
--- a/crates/rrg/src/action/execute_signed_command.rs
+++ b/crates/rrg/src/action/execute_signed_command.rs
@@ -281,16 +281,35 @@ where
             }
         }).collect::<crate::session::Result<Vec<_>>>()?;
 
-    let mut command_process = std::process::Command::new(&command_path)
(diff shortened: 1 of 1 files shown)


panhania:
For the record, this was superseded by #249.

PR #260, opened 2026-09-07
Fuzz memory actions

pmateatg:
Add fuzz targets for process memory dumping and YARA scanning backed by in-memory mock readers to ensure fast and isolated execution. 

Resolve two infinite loop issues caused by zero-length chunk sizes in YARA scanning and empty chunk reads during memory dumping found during creating the fuzzers.

Refactor scan_regions to mirror the structure of dump_regions, making memory scanning directly testable in RAM.

### DIFF
--- a/crates/rrg/src/action/scan_memory_yara.rs
+++ b/crates/rrg/src/action/scan_memory_yara.rs
@@ -193,7 +193,7 @@ impl From<Match> for proto::Match {
 }
 
 #[derive(Debug)]
-enum Error {
+pub enum Error {
     /// Failed to open the process' memory for reading.
     OpenProcessMemory(std::io::Error),
     /// There was an error (e.g. a timeout) when scanning memory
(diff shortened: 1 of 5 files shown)


crates/rrg/src/action/scan_memory_yara.rs
```diff
@@ -260,27 +280,84 @@ impl crate::response::Item for Item {
...
-        let remaining = region.end_address() - offset;
-        let length = remaining.min(chunk_size + chunk_overlap);
-        // Any process will most likely have at least one region
-        // that cannot be read successfully, so there's no point
-        // in reporting an error to the user if that happens,
-        // just ignore it and continue.
-        if let Ok(buf) = memory.read_chunk(offset, length) {
-            scanner.scan(offset as usize, &buf).map_err(Error::Scan)?;
+    timeout: Option<Duration>,
+    max_matches_per_pattern: Option<usize>,
```
panhania: I feel like even the original function was at the limit of its parameter count and this one is just way too much. With half of parameters being just numbers, the calls in tests are especially unreadable.

At this point, the abstraction level of this function also feels pretty arbitrary (it seems like it is just "the other half of `handle`" now) but I will let @CRefice decide what should be the best course of action here.

crates/rrg/src/action/scan_memory_yara.rs
```diff
@@ -100,7 +111,16 @@ impl crate::request::Args for Args {
                 skip_executable_regions: proto.skip_executable_regions,
                 skip_readonly_regions: proto.skip_readonly_regions,
             },
-            chunk_size: proto.chunk_size.unwrap_or(DEFAULT_CHUNK_SIZE),
+            chunk_size: match proto.chunk_size {
```
CRefice: Might be worth turning `chunk_size` into a [NonZeroU64](https://doc.rust-lang.org/stable/std/num/type.NonZeroU64.html#method.new) to maintain the non-zero invariant in the type system instead of checking for it later in `scan_region`

crates/rrg/src/action/scan_memory_yara.rs
CRefice: +1. I understand that this made injecting a fake `MemoryReader` impl easier and arguably is more in line with `dump_regions`, but this function is now doing way too much.

I can think of two approaches off the top of my head:
1. Move only the `for region in regions` loop inside this function, and keep all the scanner setup and session handling in the `handle` function (and replicate at least part of it in the fuzzer). Upside: smallest diff to existing logic; downside: lowest coverage for the fuzzer.
2. Keep this function as it is, and instead move all of the logic in `handle` to a new function that accepts `Args` as well as some kind of factory function for `MemoryReader`, that the fuzzer would then call with a fake impl.  This would mean almost all of the action logic is covered by the fuzzer, at the cost of maybe some slight breaking of abstraction boundaries.

I might lean slightly for option 1, but I'd be fine with either.

CRefice (changes requested):

pmateatg:
Thanks, I'm leaning towards Option 1 too. I'd like to keep the changes out of the core for fuzzing as much as possible. This change looked acceptable, because dump_regions already does it this way, but I'll see if it can be done better.

For what it's worth, we could break abstractions with exposing functions for fuzzing via [#[cfg(fuzzing)]](
https://github.com/rust-fuzz/book/blob/master/src/cargo-fuzz/guide.md#cfgfuzzing) as well, but I don't want to cross that line. We can and already use quickcheck in those cases where it matters and exposure doesn't make sense.

CRefice (approved):

